PR #13, opened 2019-04-12
Coordinate support

Kokaiinum:
Well, here it is. Tested it with your converted acc hair.


DeathWeasel1337:
Very nice, thank you!

PR #27, opened 2019-12-04
add AI_StudioSceneLoadedSound, add DragAndDrop support to StudioSceneLoadedSound

GeBo1:
If you're interested, this adds AI_StudioSceneLoadedSound, and patches support in for DragAndDrop (soft dependency so plugin should work fine if it's not installed).


GeBo1:
I think I broke something when I was cleaning out my log messages. I'll debug it and update this PR.

PR #52, opened 2020-08-05
Remove redundant null checks

UselessToucan:


DeathWeasel1337:
The null checks within UpdateSkinOverlay are not redundant, loading the default male character in Honey Select 2 for example will result in nullref here. Null checks are one of the cheapest possible operations and even running hundreds of thousands of them shows no statistically significant difference in run times, with the non-null check times coming out slightly slower as often as not. That said, I don't think it's worth my time or yours on hunting down possible redundant null checks, much less so on testing afterwards.

PR #63, opened 2020-11-09
Subtitles: Updated HS2 hooks

GeBo1:
Now works with HS2 concierge, tutorial, and intro scenes


GeBo1:
Updated XML files incoming

PR #73, opened 2021-03-06
Update DynamicBoneEditor.Controller.cs

Madevil:
need this change in line #104
or the coordinate always point to the one when card saved


DeathWeasel1337:
Thanks!

PR #75, opened 2021-03-15
Fix for reloading in Koikatsu H

jalil49:
I posted some videos in #mod-programming of Koikatsu discord. Probably needs some testing to see if it works in other games, but it should be fine with #if KK conditions

[video of ME fix how it behaves currently is the 2nd video above
](https://discord.com/channels/447114928785063977/476358846764548137/821131692893798432)


DeathWeasel1337:
Discussed it on Discord and came to a more complete solution, should be fixed as of 3.0.4

PR #85, opened 2021-05-19
Fix HS2 monochrome issue.

KissHug:
I cannot figure out what is exact problem, but I suffered monochrome does not working properly and I made a suppress code.
It is not efficient but working.


DeathWeasel1337:
Can you explain the problem? I might be able to come up with a more efficient solution. Does it happen in H mode or Studio? I don't play much HS2 so I haven't experienced it.

KissHug:
Maybe the timing is problem. I logged how many objects being hidden from `UpdateMonochromeBody()`, it always 0. Then, fixed if give some delay to it.

PR #98, opened 2021-09-12
Update ph_shaders.unity3d

AgiShark:
Replaced Standard shader and Standard_culloff shader in ph_shaders.unity3d embedded in Material Editor.dll to contain enough shader variants.

The `ph_shaders.unity3d` file embedded in PH_MaterialEditor.dll contains shaders collected from vanilla game files.
Among them, the `standard` shader does not include `shader variants` that enable even basic keywords such as _NORMALMAP,_ALPHAPREMULTIPLY_ON, so if you swap to that standard shader, Immediately, the mods will drop normal map textures.
`_NORMALMAP` is the most notable one, and other keywords (such as _EMISSION,DETAIL_MULX2_...) also don't work at all.
So, PH users are using the `Standard_555` ME shader mod as a replacement for the Standard shader, but there is another issue.

When you swap the shader of PH mods using the standard shader, press the `Reset` button in ME, the mods now use the standard shader included in ph_shaders.unity3d, As a result, the mods' normal map textures become unavailable.
So users have to reload the mods to restore them.

For the above reasons, I thought it would be better to replace the Standard shader and Standard_culloff shader, which users use most often when making mods.

https://youtu.be/XNVv0_bNLCc

Thank you for maintaining such a great plugin.


ManlyMarco:
Looks good, thank you for the PR.

PR #126, opened 2022-02-26
Fix random button in Sunshine FreeH VR

ASDAFL:
Added pre-processor tags where needed for VR differences between KK and KKS.


DeathWeasel1337:
Thanks

PR #132, opened 2022-04-15
Created Partial High Poly mode

### LINKED ISSUE #104: [KKS] ForceHighPoly issues
> steps to reproduce:
Get the island girl card, make her fully naked, save and use that as default
Start the game, go to her and observe that she looks fine
Skip some day time from morning to next one for example
Go to her and now she has a dick
the BepInEx folder https://drive.google.com/file/d/1ZNdRv_5d-NAiPT-W4VX-uc8EGV_-3ugN/view?usp=sharing
the UserData folder https://drive.google.com/file/d/15yqr5bdWvxu-yUalYTqrP33KCLK-FNk6/view?usp=sharing
with those to reproduce just load the 3rd save (row 2 column 1)
then walk to her, skip time and walk back
happens without uncensor selector

jalil49:
Enables high poly on a per-character basis
It is enabled when there is a clothing piece that lacks a low poly counterpart.
By extension preventing errors of people who lack certain pieces throwing errors

Enabled Cannot be renamed

external plugin uses enabled

Appears to fix #104


src/ForceHighPoly.Core/ForceHighPoly.cs
```diff
@@ -18,13 +17,15 @@ public partial class ForceHighPoly : BaseUnityPlugin
         public const string Version = "1.2.2";
 
         public static ConfigEntry<bool> Enabled { get; private set; }
+        public static ConfigEntry<bool> PartialPoly { get; private set; }
 
         internal void Main()
         {
             var hasEnoughRam = KKAPI.Utilities.MemoryInfo.GetCurrentStatus().ullTotalPhys > 16L * 1000L * 1000L * 1000L; // At least 16GB
             Enabled = Config.Bind("Config", "High poly mode", hasEnoughRam, "Whether or not to load high poly assets. Improves quality of characters in roaming mode and fixes some modded items not appearing.\nMay require exiting to main menu to take effect. This option has high memory requirements, at least 8GB of RAM is recommended (with ~10 characters).");
+            PartialPoly = Config.Bind("Config", "Partial High poly mode", true, "Use High Poly Assets on Characters who wear clothing that lack low poly models and prevents the resulting errors.\nMay require exiting to main menu to take effect. This option has Lower memory requirements, at least 8GB of RAM is recommended (with ~10 characters).");
```
ManlyMarco, line 26: The partial setting is confusing, if I turn the 1st setting off and the second on does it load partially or not at all? Instead of adding a new setting it might be better to change type of the 1st setting to an enum that has None, Full, and Partial values, with partial being the default.

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
```diff
@@ -1,36 +1,224 @@
...
+            /// Might not be neccessary because of the first line of the above method, put probably also cheaper than the previous method
+            /// Equivilant to last method results of patching AssetBundleManager and trimming _low
+            /// </summary>
+            /// <param name="__result"></param>
+            [HarmonyPostfix, HarmonyPatch(typeof(ChaInfo), nameof(ChaInfo.hiPoly), MethodType.Getter)]
+            private static void HiPolyPostfix(ref bool __result)
+            {
+                __result = __result || Enabled.Value;
+            }
+            #region Transpile code that works, but animation locks
```
ManlyMarco, line 94: Probably should be removed before merging. If it's possibly going to be useful in the future it's better to keep it in a branch or a discussion.

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
```diff
@@ -1,36 +1,224 @@
...
+                            continue;
+                        }
+                        var manifestName = lib.GetInfo(ChaListDefine.KeyType.MainManifest);
+                        var assetBundleName = lib.GetInfo(ChaListDefine.KeyType.MainAB);
+#if KK
+                        Singleton<Manager.Character>.Instance.AddLoadAssetBundle(assetBundleName, manifestName);
+#elif KKS
+                        Manager.Character.AddLoadAssetBundle(assetBundleName, manifestName);
 #endif
+                        if (!CommonLib.LoadAsset<UnityEngine.GameObject>(assetBundleName, lib.GetInfo(ChaListDefine.KeyType.MainData) + "_low", false, manifestName) && CommonLib.LoadAsset<UnityEngine.GameObject>(assetBundleName, highAssetName, false, manifestName))
```
ManlyMarco, line 75: Does this get reused? It seems like this would slow down the load process, and if only one piece of clothing is missing a low poly version then others will get their low and high poly versions loaded. I think there's a method for checking if an asset exists, maybe that can be used?

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
```diff
@@ -1,36 +1,224 @@
...
-            private static void LoadAssetAsyncPrefix(ref string assetName) => RemoveLow(ref assetName);
-
             /// <summary>
-            /// Strip "_low" from the end of an asset name
+            /// Test all coordiantes parts to check if a low poly doesn't exist.
+            /// if low poly doesnt exist for an item set HiPoly to true and exit;
             /// </summary>
-            private static void RemoveLow(ref string assetName)
+            /// <param name="__instance"></param>
+            [HarmonyPostfix, HarmonyPatch(typeof(ChaControl), nameof(ChaControl.Initialize))]
```
ManlyMarco, line 14: Does this fire when the character changes clothes on the main map? e.g. in dressing room. If not, then the old value would be used which might be wrong.

ManlyMarco (changes requested):

src/ForceHighPoly.Core/ForceHighPoly.cs
jalil49: The full version will take priority, but indeed a Enum works

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
jalil49: This is postfixed to ChaControl initialization so it's before the character gets loaded and doesn't get caught up having to interact with any Async patches.

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
jalil49: Fired upon character creation, since game currently hates reloading the character

src/ForceHighPoly.Core/ForceHighPoly.Hooks.cs
jalil49: Tried looking into this, but looking for the _low asset's existence seems to be approximately equal amount of work and needing new code

ManlyMarco (approved):

src/ForceHighPoly.Core/ForceHighPoly.cs
```diff
@@ -16,15 +15,50 @@ public partial class ForceHighPoly : BaseUnityPlugin
...
+            PolySetting.SettingChanged += PolySetting_SettingChanged;
+            Enabled = Config.Bind("Config", "High poly mode", hasEnoughRam, new ConfigDescription("Original Setting: kept due to externnal dependency\nWhether or not to load high poly assets. Improves quality of characters in roaming mode and fixes some modded items not appearing.\nMay require exiting to main menu to take effect. This option has high memory requirements, at least 8GB of RAM is recommended (with ~10 characters).", null, new ConfigurationManagerAttributes() { Browsable = false }));
```
ManlyMarco, line 25:
```suggestion
            PolySetting = Config.Bind("Config", "Set High Poly Mode State", hasEnoughRam ? PolyMode.Full : PolyMode.Partial, "Whether or not to load high poly models. Improves quality of characters in roaming mode and fixes some modded items not appearing.\nMay require exiting to main menu to take effect.\nNone: Original behavior - Some characters may throw errors and walk around with glitched clothing.\nPartial: Use high poly models only if necessary. Higher resource usage and longer load times when using some modded clothes.\nFull: Always use high poly models. Best quality but very high resource usage. At least 8GB of RAM is recommended with ~10 characters. Load times might take up to 3x as long.");
            PolySetting.SettingChanged += PolySetting_SettingChanged;
            Enabled = Config.Bind("Config", "High poly mode", hasEnoughRam, new ConfigDescription("Original Setting: kept due to externnal dependency\nWhether or not to load high poly assets. Improves quality of characters in roaming mode and fixes some modded items not appearing.\nMay require exiting to main menu to take effect. This option has high memory requirements, at least 8GB of RAM is recommended (with ~10 characters).", null, new ConfigurationManagerAttributes() { Browsable = false }));
```

PR #153, opened 2022-07-02
Add a color picker for color properties of PH_ME

AgiShark:
eg

I just made it work on my box. But I have no coding background.
If you find something I've missed or could improve on, I'd appreciate it if you let me know.
Users of the latest PH repack probably need to update the Bepinex and Bepisplugins.

Thanks


DeathWeasel1337:
Thank you!

Issue #141, opened 2022-04-29
[HS2]StudioWindowResize Outfit list overflowing


BullOnMars:
Title, in studio, the outfit list is overflowing in HS2, other games not tested.
Plus UI incompatibility with Browser Folders 2.6.2, list is overlapping with Browser Folders UI, since Browser Folders UI cannot be moved.

ManlyMarco:
Confirmed.
Overflowing is caused by [this](https://github.com/IllusionMods/KK_Plugins/blob/master/src/StudioWindowResize.Core/StudioWindowResizePlugin.cs#L119) line. Character lists need to be checked as well, might need to change over to the Strict version of the method.
Overlapping with Browser Folders is known, it needs to be fixed on the Browser Folders side by letting the windows be resized.

PR #157, opened 2022-07-14
Fix Costume list overflow & add resize for Image Board list

kkykkykky:
In regard to #141. Tested in all but KK. Mines too old 😛.
Thanks!


DeathWeasel1337:
Thanks

PR #160, opened 2022-08-03
Fix conflict between MoreOutfits and CharaStateX

GeBo1:
With CharaStateX installed, current version will change outfits on all selected characters to outfit worn by first selected character just by opening the "Current State" dialog.  Solved by skipping call to `OnClickCosType` in cases where it would be a no-op.


ManlyMarco (approved):
Looks good to me.

PR #165, opened 2022-11-25
Reduced radius of AI/HS2 colliders

Avvilimento:
Reduced the radius of all AI/HS2 colliders to 0.05. This assists with visible distortion in larger breasts in H scenes when using animations that have the arms close to the body (eg Standing, Standing Missionary/Lifting, Table/Desk Missionary, and Femdom Cowgirl, all in "low" variant) but does not correct it entirely (particularly past 90 size). Some clipping is introduced. NB This has no effect on AI as the existing colliders are active with this plugin and they have a larger radius.

The issue preventing a more complete solution appears to stem from the H scene skeleton having the arm bones in a different place to everywhere else in the game (and studio). As I lack the tools to visualise them in-scene I can only speculate, but it appears as though they are further forward and towards the centre line than they should be. As the bones rotate for some animations (particularly the forearms), offsets on the colliders cannot be used to reliably correct for the disparity.

I have included remarks containing the default AI colliders for posterity. If the skeleton issue can be identified and corrected these would be better values than the ones I have set.

### DIFF
--- a/src/Colliders.Core/Core.Colliders.Constants.cs
+++ b/src/Colliders.Core/Core.Colliders.Constants.cs
@@ -7,14 +7,23 @@ namespace KK_Plugins
     {
 #if AI || HS2
         internal static ColliderData FloorColliderData = new ColliderData("cf_J_Root", 100f, 0f, new Vector3(0, -100.01f, 0f));
+/*      AIS default colliders for future reference:
+            "cf_J_Hand_s_L", 0.3f, 1f, new Vector3(-0.25f, 0f, 0f)
+            "cf_J_Hand_s_R", 0.3f, 1f, new Vector3(0.25f, 0f, 0f)
+            "cf_J_ArmLow01_s_L", 0.4f, 1.5f, new Vector3(-0.35f, 0f, 0f)
+            "cf_J_ArmLow01_s_R", 0.4f, 1.5f, new Vector3(0.35f, 0f, 0f)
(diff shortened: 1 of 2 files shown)


Avvilimento:
New idea - one which I'm more pleased with.

I've set the colliders for AI/HS2 to match the default colliders within AI. This feels like a better long term strategy as it means less faffing about trying to disable the internal AI colliders. I've subsequently told it not to add the superfluous arm colliders to AI (the internal colliders still work fine - including the leg colliders that are present but seemingly inactive in HS2).

Now instead of reducing the size of the arm colliders I've instead reduced the multiplier on the breast colliders. There are two values assigned to this by default - 1.0f and 0.8f. I've been unable to establish what, if anything the second value does so have left it alone, but reducing the first value to 0.6f has much the same effect as the previous arm bone reduction with the added benefit of also working within AI. Values between 0.5f and 0.7f seem to work the best, with more clipping lower and more distortion higher.

Note that this introduces a small amount of distortion during the conversation/intro scenes in HS2 during poses where the girl's arm is raised. This can possibly be reduced by further tweaking but has no impact on the actual scenes themselves.

Avvilimento:
Final change. Disabled (commented - left for posterity) the upper forearm colliders in HS2 to deal with _most_ of the distortion at the conversation window. It still has some, just it isn't as egregious. Seems to have zero impact in-scene, which is to be expected as this is mostly covered by the other colliders.

I also figured out what the second value is for. It appears to be related to the "length" of the breast collider. Oddly enough 0.8f is about as good as this gets.

ManlyMarco (approved):

PR #172, opened 2023-03-04
Favorite items list

0fsk:
I've been using the blacklist function as the only way to find back and compare individual items among the thousands available, so I tried my hand at making a proper separated favorite system. 
I extended the blacklist plugin instead of copy-pasting its code into a new one, since the logic is basically the same, should the plugin be renamed? I didn't initially, and made sure to keep the same function signatures that were initially available, since I don't know about the ecosystem and if other stuff can depend on this plugin.
Also, I saw grouping vanilla items is disabled because they don't have a GUID, is it possible to use their standard IDs, or are those not fixed?

### DIFF
--- a/src/ItemBlacklist.Core/ItemBlacklist.UI.cs
+++ b/src/ItemBlacklist.Core/ItemBlacklist.UI.cs
@@ -58,17 +58,17 @@ namespace KK_Plugins
             scrollRect.GetComponent<Image>().color = RowColor;
 
             {
-                var contentItem = UIUtility.CreatePanel("BlacklistContent", scrollRect.content.transform);
+                var contentItem = UIUtility.CreatePanel("FavoriteContent", scrollRect.content.transform);
                 contentItem.gameObject.AddComponent<LayoutElement>().preferredHeight = PanelHeight;
                 contentItem.gameObject.AddComponent<Mask>();
                 contentItem.color = RowColor;
 
-                var itemPanel = UIUtility.CreatePanel("BlacklistPanel", contentItem.transform);
+                var itemPanel = UIUtility.CreatePanel("FavoritePanel", contentItem.transform);
                 itemPanel.color = RowColor;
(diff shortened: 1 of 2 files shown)


ManlyMarco:
Vanilla items always have the same IDs, so you can treat them as a mod with a null/empty guid.

src/ItemBlacklist.Core/ItemBlacklist.UI.cs
```diff
@@ -55,6 +57,42 @@ protected void InitUI()
...
+                contentItem.gameObject.AddComponent<LayoutElement>().preferredHeight = PanelHeight;
+                contentItem.gameObject.AddComponent<Mask>();
+                contentItem.color = RowColor;
+
+                var itemPanel = UIUtility.CreatePanel("BlacklistModPanel", contentItem.transform);
+                itemPanel.color = RowColor;
+                itemPanel.gameObject.AddComponent<CanvasGroup>();
+                itemPanel.gameObject.AddComponent<HorizontalLayoutGroup>().padding = Padding;
+
+                FavoriteModButton = UIUtility.CreateButton("BlacklistModButton", itemPanel.transform, "Favorite all items from this mod");
```
ManlyMarco, line 89: Why call these controls "Blacklist..." and not "Favorite..."?

ManlyMarco (commented):
The changes look good to me. They shouldn't conflict with any other plugins. Please let me know if you're going to make the vanilla item change you mentioned or if this is already ready to merge.

src/ItemBlacklist.Core/ItemBlacklist.UI.cs
0fsk: I wasn't sure if the string had a special function, fixed now

0fsk:
I made the changes for vanilla items support, I'm OK with merging now.

ManlyMarco (approved):

PR #180, opened 2023-05-14
[ME] Add shader keyword support.

LumineTheMeme:
I've been commissioned by an HS2 modder to add shader keyword support to MaterialEditor. I've verified that this works in both HS2 and KK.

To use it, all one needs to do is find the multi_compile shader keywords in the shader source and add them to the shader manifest using Type="Keyword"

image

image

I also incremented ME's version number from 3.1.22 to 3.1.23


ManlyMarco (approved):
Looks good, thank you for the PR!

PR #184, opened 2023-08-09
[ME] Add more ShadowCasting toggle hotkeys

RikkiBalboa:
Adds two new toggle shortcuts to set shadow casting mode to "Shadows Only" or "Two Sided". Also handles the shadowcaster renderer settings to keep consistent shadows between on and shadows only/two sided states


RikkiBalboa:
Some explanation for these force-pushes. My git was still configured for my normal account and I'd rather not have that linked and I tried to fix that 😅. It still seems to remain in the history here, but I guess the main commit should now display this account.

ManlyMarco (approved):

ManlyMarco:
Delete your fork after this is merged co get rid of the old commits.

PR #189, opened 2023-10-02
[ME] New feature: Recalculate Normals

NiggoJaecha:
Added a new RendererProperty that is available for SkinnedMeshRenderers which upon activating will recalculate the normals of the shared mesh based on the baked shape and not the original shape.
Turning the property off after it's been activated requires a reload of the according asset (similar to resetting textures).


NiggoJaecha:
Note:
Visual Studio changed the inconsistent new line characters in two of the .cs file which is why there is such a huge +/- stat on this PR.

ManlyMarco (approved):

PR #190, opened 2023-10-06
[ME] Improve search with wildcards, material property filtering and filter persistence

RikkiBalboa:
wildcard support: translate "*" and "?" to regex (".*" and ".") in the background for search.
property filtering: Searches prefixed with an underscore will be used to filter for properties instead, without it will behave as normal
filter persistence: Added a setting to allow the filter text to persist between windows. Allows for easier mass editing between objects

Added it all in one go since it's all search related

### DIFF
--- a/src/MaterialEditor.Base/UI/UI.cs
+++ b/src/MaterialEditor.Base/UI/UI.cs
@@ -363,7 +363,7 @@ namespace MaterialEditorAPI
                     string propertyName = property.Key;
                     if (Instance.CheckBlacklist(materialName, propertyName)) continue;
 
-                    bool showProperty = filterListProperties.Count == 0 ? true : false;
+                    bool showProperty = filterListProperties.Count == 0;
                     foreach (string filterWord in filterListProperties)
                         if (WildCardSearch(propertyName, filterWord))
                         {
--- a/src/MaterialEditor.Base/UI/UI.cs
+++ b/src/MaterialEditor.Base/UI/UI.cs
@@ -4,6 +4,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using UILib;
 using UnityEngine;
 using UnityEngine.UI;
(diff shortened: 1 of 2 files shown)


BullOnMars:
Extremely useful features.

RikkiBalboa:
I'm trying to add a checkbox next to the filter button now, to easily enable and disable the search persistence feature. Some people gave the feedback that's it very useful, but can also get in the way when not needed.

src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -352,9 +371,17 @@ protected void PopulateList(GameObject go, object data, string filter = "")
...
+                        if (WildCardSearch(propertyName, filterWord))
+                        {
+                            showProperty = true;
+                            break;
+                        }
+
+                    if (showProperty && mat.HasProperty($"_{propertyName}"))
                     {
-                        if (mat.HasProperty($"_{propertyName}"))
+                        if (property.Value.Type == ShaderPropertyType.Texture)
```
ManlyMarco, line 384: Why was the order of those checks changed? The `.HasProperty` call is much more expensive than comparing the Type which is an enum and should only be done if necessary.

Also, the Keyword type is not filtered, is this intentional?

RikkiBalboa: ah, that explains why the if check was done multiple time instead of once at the top. I will change it back.

The keyword type I just forgot about tbh, so I'll add that too

RikkiBalboa:
I switched the if checks around again and added filtering for the keywords too.
I also added the checkbox for the persistence and it seems to scale fine with the window. Right now it's still coupled to the original config entry I added, but could also very easily be removed as a config entry if that's better.

ManlyMarco:
Since now the check affected all possibilities it was possible to simplify the changes a lot by checking if filters match once at the start and skip to next loop iteration if they don't.

ManlyMarco (approved):

RikkiBalboa:
That's indeed a lot cleaner. It was sort of my attempt with swapping those ifs around at first, but didn't consider the performance.
And thanks for the merge!

PR #191, opened 2023-10-12
[ME] Timeline compatibility

RikkiBalboa:
Adds timeline compatibility to ME.

Because of how interoperables are added to timeline, I had to implement some way of selecting certain properties in the UI to let timeline know what to actually target. I did this by adding a little button next to the properties that you can timeline. These buttons are never shown in maker and need to be enabled in the config to show, since their use is a little confusing I think (but I had very little space to work with)
image

Another alternative I explored was adding checkboxes/toggles that act as radio buttons instead of normal buttons. That way you can see what is selected. But adding that seemed to require some hacky workarounds, and made the UI feel sluggish when toggling them, since it required a full UI refresh. As a compromise the button does show a message of what is selected and making it clear what the button did.

I also added support for adding texture changes to timeline. But right now they are saved as a base64 string in the timeline xml. It would maybe be possible to save these in the internal dict ME already keeps track of and just save a key to that instead. But I couldn't figure out how to get that working. So in  the meantime the interoperable has a big warning for its use (it explodes scene file sizes)

I'm also making the assumption that characters are saved the same way across all games within a scene. Namely that every character is added under a unique gameobject name (e.g. `chaf_00`), unlike normal items which can have a duplicate name. (this is the case for KK/KKS, but I don't have the other games, so I can't test this)
The reason being the fact that figuring out and saving/loading which game object a operable belonged to would require a lot more more, instead of just saving its path and finding it again.


ManlyMarco:
I'm not too keen on merging this because of the scene file size issue. It should be possible to only store the texture ID, main issue will be detecting when a texture is no longer used but even if that's not perfect it'll be way better than the current setup.

I think the character object names are always unique but I'm not completely sure, especially when other plugins are involved. Either way, doesn't this cause issues when trying to add keyframes for studio items? Check how other plugins handle this I guess.

RikkiBalboa:
I will see what I can do about storing the textures with ME instead. But disabling it completely shouldn't be a big loss since I don't expect it to be used often, mostly added purely based on the technicality that I could.

And keyframes with studio items aren't a problem. Timeline gives you an ObjectCtrInfo with which I can check if the object is an item or a character, and handle retrieving of the game object accordingly.
https://github.com/RikkiBalboa/KK_Plugins/blob/1358e62d973900015928a459f94301fcddc97e3a/src/MaterialEditor.Core.Studio/TimelineCompatibility.cs#L308C1-L314C29

Render Editor (which can also add material edits to timeline) can handle this differently because of the way the UI is structured. In RE you first select a renderer, then a material and then your property. With all that info it's easy to retrieve the correct material again by just querying all renderers on the given ObjectCtrInfo . But because of how ME is structured I have no easy way to figure out what renderer the material belonged to, so I have to use the game object instead.

And character object names are unique in KK/KKS, I checked that. Just not too sure if they're handled the same way in HS/AI, but I just assume that logic wouldn't have changed between all the games. But I'll ask around/see if I can test that myself to be sure.

BullOnMars:
Off topic slightly, would love you to look into and fix the biggest bug of timeline currently, blend shape morphs are discarded/deleted completely when switching a character to a character with a different head.

BullOnMars:
> I will see what I can do about storing the textures with ME instead. But disabling it completely shouldn't be a big loss since I don't expect it to be used often, mostly added purely based on the technicality that I could.
> 
> 
> 
> And keyframes with studio items aren't a problem. Timeline gives you an ObjectCtrInfo with which I can check if the object is an item or a character, and handle retrieving of the game object accordingly.
> 
> https://github.com/RikkiBalboa/KK_Plugins/blob/1358e62d973900015928a459f94301fcddc97e3a/src/MaterialEditor.Core.Studio/TimelineCompatibility.cs#L308C1-L314C29
> 
> 
> 
> Render Editor (which can also add material edits to timeline) can handle this differently because of the way the UI is structured. In RE you first select a renderer, then a material and then your property. With all that info it's easy to retrieve the correct material again by just querying all renderers on the given ObjectCtrInfo . But because of how ME is structured I have no easy way to figure out what renderer the material belonged to, so I have to use the game object instead.
> 
> 
> 
> And character object names are unique in KK/KKS, I checked that. Just not too sure if they're handled the same way in HS/AI, but I just assume that logic wouldn't have changed between all the games. But I'll ask around/see if I can test that myself to be sure.


Off topic slightly, would love you to look into and fix the biggest bug of timeline currently, blend shape morphs are discarded/deleted completely when switching a character to a character with a different head.

ManlyMarco (approved):

PR #193, opened 2023-10-19
[ME] Don't trigger character card reload for coordinate change in Roaming

mosirnik:
As stated in the comment, a character card reload is only necessary in an H scene. In the Roaming scene, this causes a noticeable stutter and should be avoided.

### DIFF
--- a/src/MaterialEditor.Core/Core.MaterialEditor.CharaController.cs
+++ b/src/MaterialEditor.Core/Core.MaterialEditor.CharaController.cs
@@ -855,7 +855,7 @@ namespace KK_Plugins.MaterialEditor
         internal void CoordinateChangedEvent()
         {
             //In H if a coordinate is loaded the data will be overwritten. When switching coordinates the ExtSave data must be reloaded to restore the original.
-            if (KoikatuAPI.GetCurrentGameMode() == GameMode.MainGame)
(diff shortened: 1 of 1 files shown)


ManlyMarco (commented):
Characters do change their coordinates on the map, is that not affected by this change? This will also cause this code to run in FreeH where it didn't previously.

mosirnik:
> Characters do change their coordinates on the map, is that not affected by this change?

It should not be affected. On the map, characters do transition between different coordinate slots, but the contents of each slot are not modified, so a full card reload is not required.

If I understand correctly, this reload logic is needed in H because an H scene allows the character's current coordinate slot to be overwritten by loading a coordinate card from the disk. Unlike in Maker, such a change is not persistent, meaning it is lost as soon as a different slot is selected. The call to `LoadCharacterExtSaveData` is required precisely for this situation, in order to "forget" about any previous changes made to that coordinate slot.

A related discussion on Discord starts here: https://discord.com/channels/447114928785063977/476358846764548137/821086849924988958

> This will also cause this code to run in FreeH where it didn't previously.

I don't think so. `KoikatuAPI.GetCurrentGameMode()` returns `MainGame` when called in FreeH, so the code always ran in FreeH.

ManlyMarco (approved):
Sounds good to me. Thank you for the PR!

PR #198, opened 2023-10-28
[ME] Add texture changing to timeline support

RikkiBalboa:
Was able to make the changing of textures in timeline work without touching timeline itself. I just had to move the getting of the actual texture to when the value is actually being set (`interpolateBefore`), instead of doing it on load (`readValueFromXml`).

Textures are saved in ME's texture dictionary and are prevented from being purged so long as keyframes exist that reference the ID.

I did have to make a private method public, and add some public methods to access a private variable. Couldn't think of a better way to handle this, since I need direct access to the texture ID's in order to save them in timeline.


ManlyMarco (commented):
The changes look good, although I'd rather not add new API surface. Can't the methods be made internal instead?

RikkiBalboa:
I wasnt happpy with that either, but I was under the impression that internal was for the namespace, not the whole assembly, so that's why. My c# is still a little rusty.
I fixed it now!

ManlyMarco (approved):

Issue #185, opened 2023-08-13
 [HS2] Enhancement Request: Remember and Display Last Selected Item in Shader Dropdown List


i486:
**Issue Description:**
Currently, when using the MaterialEditor plugin to select shaders from the dropdown list, the list resets its position to the top each time a new shader is selected. This behavior requires users to scroll down and locate their previously selected shader in order to choose the next one. This process can be time-consuming and cumbersome, especially when working with a long list of shaders. 

**Steps to Reproduce:**
1. Open the MaterialEditor plugin in the game.
2. Click on the shader dropdown list.
3. Select a shader from the list.
4. Click on the shader dropdown list again.

**Current Behavior:**
Upon clicking the dropdown list again, the list resets its position to the top, requiring the user to manually scroll down to locate the previously selected shader and choose the next one.

**Desired Behavior:**
It would greatly enhance the user experience if the MaterialEditor plugin could remember and display the last selected shader upon clicking the dropdown list. This means that after a user selects a shader, the next time they open the dropdown list, it should automatically scroll to the position of the last selected shader, keeping it visible without requiring the user to scroll through the entire list again.

**Additional Notes:**
If making it the default behavior is deemed too significant a change, then providing an option within the plugin settings would be great.

**Thank you for considering this request.**

PR #199, opened 2023-10-31
[ME] Initial display of the shader currently selected in the drop-down

takahiro0327:
The currently selected shader will be initially displayed in the center of the dropdown.

I implemented the following issues.
https://github.com/IllusionMods/KK_Plugins/issues/185

The following use cases have been useful to me
- When switching to a similar shader in the vicinity on the list
- When I miss-selected a shader

The operation was checked at KK, KKS, and HS2 studios.
If you can use it, please merge it.


ManlyMarco (approved):
Looks good, thank you for the PR!

PR #202, opened 2023-11-11
[ME] Disable Cache and Cache-clear methods

NiggoJaecha:
Adds a plugin config entry for disabling the cache and public methods to clear the entire or a specific game object from the Renderer cache. 
This aims to solve issues that are related to the caching, like the multi mesh issue with ObjImport.
Users can disable the cache to see if it fixes whatever issue they are having, or call the cache clear method with RUE, modders can specifically clear the cache for their GameObjects if they change the renderers.

### DIFF
--- a/src/MaterialEditor.Core/Core.MaterialEditor.Hooks.cs
+++ b/src/MaterialEditor.Core/Core.MaterialEditor.Hooks.cs
@@ -50,10 +50,17 @@ namespace KK_Plugins.MaterialEditor
         }
 
         private static readonly Dictionary<GameObject, List<Renderer>> _RendererLookup = new Dictionary<GameObject, List<Renderer>>();
+
+        internal static void ClearCache(GameObject gameObject = null)
+        {
+            if (gameObject == null) _RendererLookup.Clear();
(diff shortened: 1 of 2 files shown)


BullOnMars:
Cool

src/MaterialEditor.Core/Core.MaterialEditor.Hooks.cs
```diff
@@ -50,10 +50,17 @@ private static void GetBodyRendererList(GameObject gameObject, List<Renderer> re
         }
 
         private static readonly Dictionary<GameObject, List<Renderer>> _RendererLookup = new Dictionary<GameObject, List<Renderer>>();
+
+        internal static void ClearCache(GameObject gameObject = null)
+        {
+            if (gameObject == null) _RendererLookup.Clear();
+            else if (_RendererLookup.ContainsKey(gameObject)) _RendererLookup.Remove(gameObject);
```
ManlyMarco, line 57:
```suggestion
            else _RendererLookup.Remove(gameObject);
```
No need to check first, just remove.

ManlyMarco (approved):

PR #226, opened 2024-02-04
[C2A] Have rerigging match vanilla behavior.

LumineTheMeme:
Vanilla behavior simply targets the object root rather than reading the rend arrays. Fixes clothing items with improperly set up rend arrays. Apparently also fixes Mod Bone Implanter.


ManlyMarco:
Did you check for regressions? Do different sorts of clothes attach and move correctly in studio or h scenes?

LumineTheMeme:
Doing such testing for FreeH would be quite tedious. Seems to work just fine in Maker and Studio.

ManlyMarco (approved):

PR #229, opened 2024-02-18
Added texture manager - Share identical textures across characters

takahiro0327:
There is still room to organize the code. Like adding comments.
But before I do, I wanted to get your opinion on whether these changes are safe or not, so I made this PR.

@ManlyMarco 
I created a class to manage the textures that TextureContainer had. Is this safe?

If it matches the byte[] of the already loaded texture, it does not create a Texture, but returns the already created Texture.
This is similar to MaterialEdtor's texture management, but MaterialEditor's is treated as individual textures when the same texture is loaded for multiple characters. This change uses the same texture for all of them.

The advantages and disadvantages are as follows.

Advantages:
Load times are reduced by several tens of percent in the case of scenes with many identical characters. Memory usage is also reduced by several hundred MB.
In the scene I tested, the load time was reduced from 75 seconds to 50 seconds.

Disadvantages:
If there are plug-ins that change textures individually, it will not work correctly.
If there are no identical textures, there is no change. In fact, it slows things down slightly.
When there is a code that forgets to Dispose the TextureContainer, the texture will not disappear with UnloadUnusedAssets.

Do you think it's a good idea to include this change?
If it looks OK, I'll clean up the source. If it's a problem, close PR and be done with it.

### DIFF
--- a/src/Shared.TextureContainer/Shared.TextureContainerManager.cs
+++ b/src/Shared.TextureContainer/Shared.TextureContainerManager.cs
@@ -6,34 +6,47 @@ using UnityEngine;
 
 namespace KK_Plugins
 {
+    /// <summary>
+    /// Class that manages texture containers.
+    /// </summary>
     public sealed class TextureContainerManager
     {
-        static Dictionary<TextureHolderKey, TextureHolder> _textureHolder = new Dictionary<TextureHolderKey, TextureHolder>();
+        /// <summary>
+        /// Dictionary to hold textures
+        /// </summary>
+        static Dictionary<TextureKey, Token> _textureHolder = new Dictionary<TextureKey, Token>();
 
-        public class TextureHolderKey
+        /// <summary>
+        /// Class for using dictionary keys for information needed to generate textures
+        /// </summary>
+        internal class TextureKey
         {
             public readonly byte[] data;
-            public readonly uint dataHash;
+            public readonly int hash;
             public readonly TextureFormat format;
             public readonly bool mipmaps;
 
-            public TextureHolderKey(byte[] data, TextureFormat format = TextureFormat.ARGB32, bool mipmaps = true)
+            public TextureKey(byte[] data, TextureFormat format = TextureFormat.ARGB32, bool mipmaps = true)
             {
-                this.data = data;
                 //First part of the data is sufficient to calculate the hash.
-                dataHash = CRC32Calculator.CalculateCRC32(data, 1 << 10);
+                int hash = (int)CRC32Calculator.CalculateCRC32(data, 1 << 10);
+                hash ^= format.GetHashCode();
+                hash ^= mipmaps.GetHashCode();
+
+                this.data = data;
+                this.hash = hash;
                 this.format = format;
                 this.mipmaps = mipmaps;
             }
 
             public override bool Equals(object obj)
             {
-                if (obj is TextureHolderKey)
+                if (obj is TextureKey)
                 {
-                    var other = (TextureHolderKey)obj;
+                    var other = (TextureKey)obj;
 
                     return
-                        other.dataHash == dataHash &&
+                        other.hash == hash &&
                         other.format == format &&
                         other.mipmaps == mipmaps &&
                         other.data.SequenceEqual(data);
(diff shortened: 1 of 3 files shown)


ManlyMarco (commented):
This might cause some unintended effects like you've mentioned but I think it should be safe to do, most plugins that edit characters will replace the texture rather than modify it. If the gains are significant then it's definitely worth doing.

I would consider using a token system instead, where getting a texture results in a new token which can be Disposed after it is no longer used. If all tokens are disposed then the texture can be removed. 

I was planning on adding a global texture cache like this to KKAPI so that all textures are saved in one place, reducing card or scene size (texture data from all cards in a scene could be deduplicated and stored in scene info for example).

takahiro0327:
By token system, do you mean a system built on top of C# garbage collection?
Isn't my current implementation also a TextureContainer token?

Unlike the FindLoop case, the gains are significant.
(I didn't think that the other one would shrink by only a few seconds. I didn't recognize the profiler overhead correctly.
In many scenes, the same character appears in more than one scene.
It should be common to place characters who have changed only their positions and stances.

> I was planning on adding a global texture cache 

Did you already have a plan? Would that be implemented next week without me doing anything?

It would be nice if duplicates could be removed at save time, making the file size small.
However, I wonder if the changes and the scope of their impact would be too large.
Why not introduce it at runtime for now, and if there are no problems, extend it to file saving?

> When there is a code that forgets to Dispose the TextureContainer, the texture will not disappear with UnloadUnusedAssets.

I just noticed that this problem could be avoided if the TextureHolder has a Texture as a WeakReference.

ManlyMarco:
The global texture cache would be a completely separate API inside KKAPI that plugins would have to be updated to use, so the plugin's extended data format would have to be changed accordingly. This is not something that I am working on, just an idea for the future, so don't worry about it.

For now it would be best to not change any extended data format unnecessarily.

> By token system, do you mean a system built on top of C# garbage collection?

No, I mean something similar to how CancellationToken is used. The plugin would have to explicitly dispose the token if it is no longer being used, and if all tokens are disposed then the texture can be removed. This would make it no longer necessary to check if a texture is still being used at character/scene save, but will likely bring other issues.

takahiro0327:
> CancellationToken
Wow, I didn't know there was such a thing. I didn't know that.
But it looks a bit overkill. I'll go with what I have because refCount is sufficient for this application.
It's a single thread.

takahiro0327:
Placement of 20 characters, 10 of each of the two types, which is quite favorable to this modification, but greatly improves the situation.

- Before
Average: 61.66[s]
Memory usage after loading: 5.7[GB]
Peak memory usage during loading: 12[GB]
```
Scene loading completed: 63.8[s]
Scene loading completed: 61.6[s]
Scene loading completed: 60.3[s]
Scene loading completed: 60.5[s]
Scene loading completed: 62.1[s]
```

- After
Average: 36.32[s] -25.34
Memory usage after loading: 5.1[GB] -0.6
Peak memory usage during loading: 10[GB] -2
```
Scene loading completed: 37.4[s]
Scene loading completed: 35.2[s]
Scene loading completed: 34.7[s]
Scene loading completed: 39.8[s]
Scene loading completed: 34.5[s]
```

src/Shared.TextureContainer/Shared.TextureContainerManager.cs
```diff
@@ -0,0 +1,227 @@
...
+            //LoadImage automatically resizes the texture so the texture size doesn't matter here
+            Texture2D tex = new Texture2D(2, 2, format, mipmaps);
+
+            try
+            {
+                tex.LoadImage(texBytes);
+
+                RenderTexture rt = new RenderTexture(tex.width, tex.height, 0);
+                rt.useMipMap = mipmaps;
+                Graphics.Blit(tex, rt);
```
ManlyMarco, line 164: Is copying the texture into a new RenderTexture necessary, instead of simply returning `tex`?

The summary description of the method states that Texture2D is returned.

takahiro0327: The summary has been revised.

I would rather hear why from me.
Wasn't this a deliberate process?

I made the following changes locally.
```
Texture2D tex = new Texture2D(2, 2, format, mipmaps);
tex.LoadImage(texBytes);
return tex;
```

Loading times were virtually unchanged.
I looked at some scenes and I don't think there was any change in appearance.
For reasons unknown, memory usage after loading increased by about 1.5 GB at most.

I imagine that RenderTexture only has data in GPU memory, but Texture2D has textures in both regular memory and GPU memory, so the memory usage would be inflated.


The code comes from this commit.
https://github.com/IllusionMods/KK_Plugins/commit/50a5c7f7f06bbee41d278db24894c996492820dd

To speed up normal map changes ......?
There was no difference in load times in my environment, but that may be because I didn't have any characters to change normal maps.

ManlyMarco: You are right, RenderTexture exists only on the GPU, while Texture2D has a copy in main memory.

If nothing needs a Texture2D then this is better. Maybe also add a comment explaining why RenderTexture is used?

takahiro0327: Added comment.

ManlyMarco (approved):

PR #231, opened 2024-02-29
[ME] Optimised generation of normal maps

takahiro0327:
The normal map texture conversion performed in the material editor has been optimised.
The two main changes are.
- Conversion results are cached so that the same texture is not converted more than once.
- Converted textures are not converted again.

Please merge if this looks good.

The scene loads a little faster.
Before ave 34.2[s]:
```
Scene loading completed: 34.7[s]
Scene loading completed: 33.7[s]
Scene loading completed: 33.8[s]
```

After ave 33.22[s]  -0.98:
```
Scene loading completed: 32.6[s]
Scene loading completed: 33.1[s]
Scene loading completed: 33.8[s]
```

Peak memory usage during loading has been reduced from 10 GB to 5.6 GB.
image
**Transition between memory usage and number of textures during scene loading.**

[before-after.ods](https://github.com/IllusionMods/KK_Plugins/files/14448905/before-after.ods)
[Code used for measurement.zip](https://github.com/IllusionMods/KK_Plugins/files/14448937/Code.used.for.measurement.zip)


takahiro0327:
Build did not pass in API.MaterialEditor

ManlyMarco (approved):
Seems fine. `ConvertNormalMap` is no longer virtual so if anything else tries to override it then that will be broken, but I don't think it was ever used outside of the ME repository.

takahiro0327:
As far as I checked the C# code in Github, there were no external plugins that call or define ConvertNormalMap.
So I hope it is ok.
https://github.com/search?q=ConvertNormalMap++language%3AC%23&type=code&l=C%23

As for the matter of deleting virtual, if it causes any problems, it will probably be a compile-time error. I believe the impact is negligible.
I believe that there is no funky code that is overriding at IL level.
```
public bool ConvertNormalMap(ref Texture tex, string propertyName)
protected virtual Texture ConvertNormalMap( Texture src )
```

ManlyMarco:
Seems like nothing is using this part of the API so I'll merge it.

PR #234, opened 2024-03-07
[ME] Add toggle for UpdateWhenOffscreen MeshRenderer property

RikkiBalboa:
Some MeshRenderers like to suddenly dissapear when partly off-screen. This property _can_ be set in the mod itself, but some mods don't test if it's needed or not I guess. So useful to have when you do encounter such an item/accessory/clothing/etc.

Only shows on renderers that are MeshRenderers


ManlyMarco:
I believe this is only a problem if the mesh has a messed up bounding box. The mod itself should just have the bounds fixed instead of making it update off-screen, but this works fine as a fix for a broken mod.

ManlyMarco (approved):

PR #244, opened 2024-04-20
[ME] Projector support

RikkiBalboa:
Projector support for ME. Formely you had to use Render Editor to edit them in games that had projectors. Koikatsu doesn't have them yet, but I'm porting those, but then RE doesn't exist on KKS.

All projector properties are treated as material properties in the UI, and thus grouped together with them. I did it like this to make editing them a little easier (grouped together + copy/pasting edits), and since projectors can only ever have one material anyway. Since projectors can only have one material, I also hide the "Copy Material" button now whenever the provided function for that button is `null`.

image

Fow now I've made all character versions of the projector methods throw a `NotImplementedException`, since projectors don't seem to be used there at all. If the rest of the code is fine I will add those functions for completeness sake though (and in case they do exist there, though I doubt that)
  
There is also the problem of potential conflicts with RE, This is because projectors share their material instance by default, while renderer materials do not. To circumvent this you need to assign a unique copy yourself and keep track if you've already assigned a unique copy or not. RE seems to do this, and might cause conflicts.
In my testing they don't conflict for some reason though, except for the `orthographic` setting, which can't be edited in ME if it's enabled in RE

### DIFF
--- a/src/MaterialEditor.Base/MaterialAPI.cs
+++ b/src/MaterialEditor.Base/MaterialAPI.cs
@@ -460,17 +460,15 @@ namespace MaterialEditorAPI
         /// <returns>True if the value was set, false if it could not be set</returns>
         public static bool SetProjectorEnabled(GameObject gameObject, string projectorName, bool value)
         {
-            bool didSet = false;
-
             foreach (var projector in GetProjectorList(gameObject))
             {
                 if (projector.NameFormatted() == projectorName)
                 {
                     projector.enabled = value;
-                    didSet = true;
+                    return true;
                 }
             }
-            return didSet;
+            return false;
         }
 
         /// <summary>
@@ -482,17 +480,15 @@ namespace MaterialEditorAPI
         /// <returns>True if the value was set, false if it could not be set</returns>
         public static bool SetProjectorFarClipPlane(GameObject gameObject, string projectorName, float value)
         {
-            bool didSet = false;
-
             foreach(var projector in GetProjectorList(gameObject))
             {
                 if(projector.NameFormatted() == projectorName)
                 {
                     projector.farClipPlane = value;
-                    didSet = true;
+                    return true;
                 }
             }
-            return didSet;
+            return false;
         }
 
         /// <summary>
@@ -504,17 +500,15 @@ namespace MaterialEditorAPI
         /// <returns>True if the value was set, false if it could not be set</returns>
         public static bool SetProjectorNearClipPlane(GameObject gameObject, string projectorName, float value)
         {
-            bool didSet = false;
-
             foreach (var projector in GetProjectorList(gameObject))
             {
                 if (projector.NameFormatted() == projectorName)
                 {
                     projector.nearClipPlane = value;
-                    didSet = true;
+                    return true;
                 }
             }
-            return didSet;
+            return false;
         }
 
         /// <summary>
@@ -526,17 +520,15 @@ namespace MaterialEditorAPI
         /// <returns>True if the value was set, false if it could not be set</returns>
         public static bool SetProjectorFieldOfView(GameObject gameObject, string projectorName, float value)
         {
-            bool didSet = false;
-
             foreach (var projector in GetProjectorList(gameObject))
             {
                 if (projector.NameFormatted() == projectorName)
                 {
                     projector.fieldOfView = value;
-                    didSet = true;
+                    return true;
                 }
             }
-            return didSet;
+            return false;
         }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
```diff
@@ -589,15 +620,18 @@ protected override void OnObjectDeleted(ObjectCtrlInfo objectCtrlInfo)
...
-            if (visible && objectCtrlInfo is OCIItem item)
+            if (objectCtrlInfo is OCIItem item)
             {
                 var id = item.GetSceneId();
-                foreach (var property in RendererPropertyList.Where(x => x.ID == id && x.Property == RendererProperties.Enabled))
-                {
-                    MaterialAPI.SetRendererProperty(GetObjectByID(id), property.RendererName, property.Property, property.Value);
-                    // potential recalc of normals, have to test...
-                }
+                if(visible)
```
ManlyMarco, line 626: Why change this?

src/MaterialEditor.Core.Studio/Core.MaterialEditor.Studio.cs
```diff
@@ -15,6 +15,7 @@
 using UnityEngine;
 using UnityEngine.SceneManagement;
 using UnityEngine.UI;
+using static Illusion.Utils;
```
ManlyMarco, line 18: This is dangerous, different games have different stuff in here that might hide unexpected things.

ManlyMarco (commented):
Overall it seems fine, just a few concerning parts.

src/MaterialEditor.Base/UI/UI.ListEntry.cs
```diff
@@ -259,8 +259,13 @@ public void SetItem(ItemInfo item, bool force)
...
                         }
-                        MaterialCopyRemove.onClick.RemoveAllListeners();
-                        MaterialCopyRemove.onClick.AddListener(delegate { item.MaterialOnCopyRemove.Invoke(); });
+                        if (item.MaterialOnCopyRemove != null)
+                        {
+                            MaterialCopyRemove.onClick.RemoveAllListeners();
+                            MaterialCopyRemove.onClick.AddListener(delegate { item.MaterialOnCopyRemove.Invoke(); });
+                        }
+                        else
+                            MaterialCopyRemove.gameObject.SetActive(false);
```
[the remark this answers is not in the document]
RikkiBalboa: This is only ever touched when the UI is populated. And every time ME updates the UI, it repopulates the whole list, so it should never become not null suddenly.
ME actually adds all row types per row, but only shows those that should be enabled

example:
https://github.com/IllusionMods/KK_Plugins/blob/9a92e3b1ecb04a103bf58bdc39c3df1d7ef1538b/src/MaterialEditor.Base/UI/UI.ListEntry.cs#L752-L756

src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
RikkiBalboa: oops. First changed it to match a projectors enabled state with the items visible state. But I removed it to not mess anything up for those games that already had projectors and may have enabled items, but disabled projectors.

Will change it back to its original state

src/MaterialEditor.Core.Studio/Core.MaterialEditor.Studio.cs
RikkiBalboa: accidentally added when trying stuff I think. Actually unused and will remove it again

src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -706,8 +804,13 @@ private static void ExportTexture(Material mat, string property)
         public abstract void SetRendererProperty(object data, Renderer renderer, RendererProperties property, string value, GameObject gameObject);
         public abstract void RemoveRendererProperty(object data, Renderer renderer, RendererProperties property, GameObject gameObject);
 
-        public abstract void MaterialCopyEdits(object data, Material material, GameObject gameObject);
-        public abstract void MaterialPasteEdits(object data, Material material, GameObject gameObject);
```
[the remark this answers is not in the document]
RikkiBalboa: Going to add those back

RikkiBalboa:
Here are the projectors for KK/KKS
[[Rikki Balboa] Projectors v1.0.zip](https://github.com/IllusionMods/KK_Plugins/files/15049854/Rikki.Balboa.Projectors.v1.0.zip)

And those for HS/HS2/AI are included with RE
[[HS] RendererEditor.zip](https://github.com/IllusionMods/KK_Plugins/files/15049907/HS.RendererEditor.zip)
[[AI] RendererEditor.zip](https://github.com/IllusionMods/KK_Plugins/files/15049908/AI.RendererEditor.zip)
[[HS2] RendererEditor.zip](https://github.com/IllusionMods/KK_Plugins/files/15049909/HS2.RendererEditor.zip)

src/MaterialEditor.Base/MaterialAPI.cs
```diff
@@ -57,6 +86,8 @@ private static List<Material> GetObjectMaterials(GameObject gameObject, string m
                 foreach (var material in GetMaterials(gameObject, renderer))
                     if (material.NameFormatted() == materialName)
                         materials.Add(material);
+            foreach (var projector in GetProjectorList(gameObject))
```
ManlyMarco, line 89:
```suggestion

            foreach (var projector in GetProjectorList(gameObject))
```

ManlyMarco (approved):
Seems good, will merge after testing is done.

PR #246, opened 2024-04-26
[ME] Auto convert OpenGL normal maps to proper format

RikkiBalboa:
When setting a texture, it now checks if a conversion is needed or not, and if so, converts the normalmap using the new shader embedded in the .unity3d file.

This check is done by checking if the property name is one of the known normal map property names (same check as when converting gray normals to red), and if there is an alpha channel.
The red normals store information in the alpha channel, while the OpenGL ones don't. So if there is an alpha value below 1, it's safe to assume no conversion is needed

I'm personally not really a fan of how I use a build conditional to exclude this in the API, while making it in the API class though. But it uses and references stuff that is not referenced in the API, but is also used elsewhere. 
But since the `SetTexture` is the central method where the saving actually takes place, it makes sure this conversion always takes place when needed
I'm assuming that since the public facing methods stay the same, but only the implementation changes, this has no actual effect on the API.
But if you know a better way of doing this, I'm all ears


RikkiBalboa:
Set this to a draft because I found some problems when doing this on characters

RikkiBalboa:
Moved the code out of the API and modified the existing normal conversion method instead.
This conversion method was already called in the relevant chara controller methods, so I added it to the scene controller too.

src/MaterialEditor.Base/NormalMapManager.cs
```diff
@@ -161,6 +163,25 @@ private static Texture2D MakeTextureReadable(Texture tex, RenderTextureFormat rt
                 RenderTexture.ReleaseTemporary(tmp);
             }
         }
+
+        internal static bool IsUncompressedNormalMap(Texture tex)
+        {
+            Texture2D readableTex = MakeTextureReadable(tex);
```
takahiro0327, line 169: Sorry for cutting in.
Memory leak if readableTex is not released.

RikkiBalboa: Thank you! Fixed it now

ManlyMarco (approved):

PR #258, opened 2024-06-01
[ME] Drag labels to change float attached float values

RikkiBalboa:
Allows for dragging on the labels to change values, just like it works in the transform panel in studio, or the accessory transform panel in maker.

It accepts a few modifiers to change thebehaviour:
- `shift`: Speed up movement by a factor of 10
- `ctrl`: Slow down movement by a factor of 10
- `alt`: Also change a connected field an even amount (e.g. also change X when changing Y)
- `right click`: When used in combination with `alt`, keeps the current ratio between the values intact

This works for all the different float values ME shows
- Texture offset (Combines X and Y for `alt`)
- Texture scale (Combines X and Y for `alt`)
- RGBA values (Combines RGB for `alt`)
- Float sliders (Dragging the label ignores the slider min/max)

https://github.com/IllusionMods/KK_Plugins/assets/141709004/84cbb990-0223-42c7-9668-24fa77ac7afc

### DIFF
--- a/src/MaterialEditor.Base/UI/UI.ListEntry.cs
+++ b/src/MaterialEditor.Base/UI/UI.ListEntry.cs
@@ -1,4 +1,8 @@
 ﻿using System;
+#if !API
+using UniRx;
+using UniRx.Triggers;
+#endif
 using UnityEngine;
 using UnityEngine.UI;
 using static UILib.Extensions;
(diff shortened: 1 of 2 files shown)


src/MaterialEditor.Base/UI/UI.ListEntry.cs
```diff
@@ -840,6 +857,38 @@ private void ShowKeyword(bool visible = true)
...
+
+    /// <summary>
+    /// Subscribes to the on onDrag event on the given Text object, and updates the given input field(s) according to the drag
+    /// </summary>
+    /// <param name="dragText">The text that should be draggable</param>
+    /// <param name="inputField">The InputField that should be updated upon dragging</param>
+    /// <param name="pairedInputFields">This InputField will also be updated if alt is held while dragging (e.g. pairing X and Y fields)</param>
+    public void SubscribeToOnDragUpdateInput(Text dragText, InputField inputField, InputField[] pairedInputFields = null)
+        {
+#if !API
```
ManlyMarco, line 870: You can do this without UniRx by making a custom MonoBehavior for this.

ManlyMarco (approved):

PR #260, opened 2024-06-01
[ME] Enable toggle shortcuts in maker

RikkiBalboa:
Enable the toggle shortcuts in maker too. Apparently some madevil plugin has some similair mass toggle functionality as these shortcuts, so I just added them to maker too.
The main use case I got told was to mass enable renderers/shadows instead of going one by one.

The body itself is excluded while using the shortcut in maker because there are some renderers there that you usually don't want to mess with at all (That madevil plugin also ignores it I got told). In studio it will still affect the body renderers when using the shortcut.


ManlyMarco (commented):
You probably want to update the hotkey descriptions [here](https://github.com/RikkiBalboa/KK_Plugins/blob/ad0d19fe9d51d08f3900c63f9cf214b37f257a5a/src/MaterialEditor.Core/Core.MaterialEditor.cs#L147-L156) too.

ManlyMarco (approved):
All good, danke schon.

PR #261, opened 2024-06-02
[ME] WebP support

RikkiBalboa:
Reads the file signature from the bytes to determine the file format, and thus which loading method to use.
Uses code from [this repo](//https://github.com/octo-code/webp-unity3d) to read and convert WebP to a Texture2D


src/Shared.TextureContainer/WebP/Texture2DExt.cs
```diff
@@ -0,0 +1,281 @@
...
+
+                lOutputBuffer = new byte[lLength];
+                Marshal.Copy(lResult, lOutputBuffer, 0, lLength);
+            }
+            finally
+            {
+                NativeBindings.WebPSafeFree(lResult);
+            }
+
+            lPinnedArray.Free();
```
takahiro0327, line 276: I was looking at it and was a bit curious.
If an exception occurs in try-catch, Free is not called. Is that ok?

src/Shared.TextureContainer/Shared.ImageHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                byte[] magicBytes = new byte[maxMagicBytesLength];
+
+                for (int i = 0; i < maxMagicBytesLength; i += 1)
+                {
+                    magicBytes[i] = br.ReadByte();
+
+                    foreach (var kvPair in imageFormatDecoders)
+                    {
+                        if (magicBytes.StartsWith(kvPair.Key))
```
takahiro0327, line 32: Forgive me for chiming in
It's not related to WebP, and right now it's harmless... I was curious.
The number of bytes read is not taken into account. Wouldn't all data with the first 1byte of 0x00 be Avif?

How about something like this?
```
public static ImageFormat GetContentType(byte[] imageBytes)
{
    foreach( var kvPair in imageFormatDecoders )
    {
        var magicBytes = kvPair.Key;
        if (magicBytes.Length <= imageBytes.Length && imageBytes.StartsWith(magicBytes))
            return kvPair.Value;
    }
    return ImageFormat.Unrecognized;
}
```

RikkiBalboa: Thanks! Now that I take a good look, that would indeed be the case. Your code also looks a lot simpler to digest. I think this is even better:
```csharp
public static ImageFormat GetContentType(byte[] imageBytes)
{
    foreach (var kvPair in imageFormatDecoders.OrderByDescending(x => x.Key.Length))
        if (kvPair.Key.Length <= imageBytes.Length && imageBytes.StartsWith(kvPair.Key))
            return kvPair.Value;
    return ImageFormat.Unrecognized;
}
```

Now it handles larger magic number sequences first, which should solve most issues (if such clashing file formats were to ever be supported)

src/Shared.TextureContainer/WebP/Texture2DExt.cs
RikkiBalboa: Probably not, moved it to the finally too. Thanks!

src/Shared.TextureContainer/Shared.ImageHelper.cs
takahiro0327: I see. Indeed.
If the magics are compatible, I'll have to start looking at the long ones.

takahiro0327:
@RikkiBalboa 
Sorry, Can you give me the webp dll?

RikkiBalboa:
@takahiro0327 I didn't link to any external dll files. But I use the code from [webp-unity3d ](https://github.com/octo-code/webp-unity3d) to load them, which uses [libwebp-net](https://github.com/imazen/libwebp-net).

takahiro0327:
@RikkiBalboa 
DllNotFoundException until I added libwebp.dll.
I think I need to include this DLL somewhere?

```
[Error  :Modding API] System.DllNotFoundException: webp
  at (wrapper managed-to-native) WebP.Extern.NativeBindings+External.WebPGetInfo(intptr,uintptr,int&,int&)
  at WebP.Extern.NativeBindings.WebPGetInfo (System.IntPtr data, System.UIntPtr data_size, System.Int32& width, System.Int32& height) [0x00008] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at WebP.Texture2DExt.GetWebPDimensions (System.Byte[] lData, System.Int32& lWidth, System.Int32& lHeight) [0x0002d] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at WebP.Texture2DExt.CreateTexture2DFromWebP (System.Byte[] lData, System.Boolean lMipmaps, System.Boolean lLinear, WebP.Error& lError, WebP.Texture2DExt+ScalingFunction scalingFunction) [0x00013] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at KK_Plugins.TextureContainerManager.TextureFromBytes (System.Byte[] texBytes, UnityEngine.TextureFormat format, System.Boolean mipmaps) [0x000a5] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at KK_Plugins.TextureContainerManager+Token.get_Texture () [0x0001b] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at KK_Plugins.TextureContainer.get_Texture () [0x00000] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at KK_Plugins.MaterialEditor.SceneController.SetTextureWithProperty (UnityEngine.GameObject go, KK_Plugins.MaterialEditor.SceneController+MaterialTextureProperty textureProperty) [0x00041] in <9ed7f14417ec4d7ba98a8a8163adeeb6>:0 
  at (wrapper dynamic-method) KK_Plugins.MaterialEditor.SceneController.DMD<KK_Plugins.MaterialEditor.SceneController::OnSceneLoad>(KK_Plugins.MaterialEditor.SceneController,KKAPI.Studio.SaveLoad.SceneOperationKind,KKAPI.Utilities.ReadOnlyDictionary`2<int, Studio.ObjectCtrlInfo>)
  at KKAPI.Studio.SaveLoad.StudioSaveLoadApi.OnSceneBeingLoaded (KKAPI.Studio.SaveLoad.SceneOperationKind operation) [0x00053] in <6aa6736edae14b6084714249f4273e73>:0 
```

RikkiBalboa:
@takahiro0327 Most people I've tested it with seem to load it just fine. But from your situation it seems that isn't any sort of guarantee. So I'll look into something so it gets included (or at least fails more gracefully)

takahiro0327:
Please do.

I was in the minority.
I wonder if it is included in the DLL of some famous tool.

ManlyMarco:
> I wonder if it is included in the DLL of some famous tool.

It might be included in codec packs.

src/Shared.TextureContainer/Shared.TextureContainerManager.cs
```diff
@@ -151,13 +151,18 @@ public static void Release( Token holder )
...
             Texture2D tex = new Texture2D(2, 2, format, mipmaps);
 
             try
             {
-                tex.LoadImage(texBytes);
+                //Only use magic numbers for custom supported image formats. Let LoadImage handle png/jpg/unknown
+                if (imageFormat == ImageHelper.ImageFormat.WebP)
+                    tex = WebP.Texture2DExt.CreateTexture2DFromWebP(texBytes, mipmaps, false, out var error);
+                else
+                    tex.LoadImage(texBytes);
```
ManlyMarco, line 165: Maybe move all image loads to a method in `ImageHelper` so it's all self-contained?

src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -546,7 +546,12 @@ void PopulateListMaterial(Material mat, Projector projector = null)
                             };
                             textureItem.TextureOnImport = () =>
                             {
-                                OpenFileDialog.Show(OnFileAccept, "Open image", Application.dataPath, FileFilter);
+#if !API
+                                string fileFilter = KK_Plugins.ImageHelper.FileFilter;
+#else
+                                string fileFilter = ""
```
ManlyMarco, line 552: It's a pretty big regression for API builds.
```suggestion
                                string fileFilter = "Images (*.png;.jpg)|*.png;*.jpg|All files|*.*";
```

src/Shared.TextureContainer/Shared.ImageHelper.cs
```diff
@@ -21,6 +26,16 @@ public static ImageFormat GetContentType(byte[] imageBytes)
...
         }
 
+        public static void LoadTexture2DFromBytes(byte[] texBytes, ref Texture2D tex)
+        {
+            var imageFormat = GetContentType(texBytes);
+            //Only use magic numbers for custom supported image formats. Let LoadImage handle png/jpg/unknown
+            if (imageFormat == ImageFormat.WebP)
+                tex = WebP.Texture2DExt.CreateTexture2DFromWebP(texBytes, tex.mipmapCount > 1, false, out var error);
+            else
+                tex.LoadImage(texBytes);
```
ManlyMarco, line 36:
```suggestion
            {
                //LoadImage automatically resizes the texture so the texture size doesn't matter here
                Texture2D tex = new Texture2D(2, 2, format, mipmaps);
                tex.LoadImage(texBytes);
            }                
```

src/Shared.TextureContainer/Shared.ImageHelper.cs
```diff
@@ -21,6 +26,16 @@ public static ImageFormat GetContentType(byte[] imageBytes)
             return ImageFormat.Unrecognized;
         }
 
+        public static void LoadTexture2DFromBytes(byte[] texBytes, ref Texture2D tex)
+        {
+            var imageFormat = GetContentType(texBytes);
+            //Only use magic numbers for custom supported image formats. Let LoadImage handle png/jpg/unknown
+            if (imageFormat == ImageFormat.WebP)
+                tex = WebP.Texture2DExt.CreateTexture2DFromWebP(texBytes, tex.mipmapCount > 1, false, out var error);
```
ManlyMarco, line 34: Maybe log the error?

src/Shared.TextureContainer/WebP/Texture2DExt.cs
takahiro0327: I would like to mark this and another case as resolved, but there is no button.
See it as resolved.
image

src/Shared.TextureContainer/Shared.ImageHelper.cs
takahiro0327: resolved

RikkiBalboa:
@takahiro0327 Could you test if it now works for you? You need to put the newly included `libwebp.lib` next to where you install material editor

takahiro0327:
@RikkiBalboa 
I made a few modifications and it worked. By the way, my environment is Windows 10.

If I had left it as it is, it would not work because the following two things did not match.
- DllName of mono_dllmap_insert argument is “libwebp”.
- The argument of DllImport in NativeBindings.cs is “webp”.

I changed the argument of mono_dllmap_insert to “webp” and it works.

```
mono_dllmap_insert(IntPtr.Zero, dllName, null, nativeLibFileName, null);
```

```
[DllImportAttribute("webp", EntryPoint = "WebPGetDecoderVersion")]
public static extern int WebPGetDecoderVersion();
```

RikkiBalboa:
@takahiro0327 Thanks! I missed that. It should now use the proper names, while still allowing easy extension for any other file formats in the future

takahiro0327 (approved):

ManlyMarco (approved):

PR #268, opened 2024-06-14
[ME] Fix mipmap calculation in WebP

RikkiBalboa:
When mipmaps are generated, more bytes need to be allocated for the texture that is going to be generated. The original calculation had a chance to allocate too few bytes when using a non-square texture.
The new calcaution fixes that. Thanks to @takahiro0327 for the formula.

### DIFF
--- a/src/Shared.TextureContainer/WebP/Texture2DExt.cs
+++ b/src/Shared.TextureContainer/WebP/Texture2DExt.cs
@@ -1,12 +1,11 @@
 ﻿//https://github.com/octo-code/webp-unity3d
 using System;
-using System.Text;
-using System.Collections.Generic;
 using System.Runtime.InteropServices;
 
 using UnityEngine;
 
 using WebP.Extern;
+using KK_Plugins;
 
 namespace WebP
 {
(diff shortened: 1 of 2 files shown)


src/Shared.TextureContainer/WebP/Texture2DExt.cs
```diff
@@ -112,13 +111,8 @@ public static unsafe byte[] LoadRGBAFromWebP(byte[] lData, ref int lWidth, ref i
...
 				}
-				
-				// If mipmaps are requested we need to create 1/3 more memory for the mipmaps to be generated in.
-				int numBytesRequired = lWidth * lHeight * 4;
-				if (lMipmaps)
-				{
-					numBytesRequired = Mathf.CeilToInt((numBytesRequired * 4.0f) / 3.0f);
-				}
+
+				int numBytesRequired = ImageHelper.CalculateTextureBytes(lWidth, lHeight, lMipmaps);
```
ManlyMarco, line 115: Why not have CalculateTextureBytes right next to where it's used since it's used in only one place?

RikkiBalboa: Idea was that if anything else would need it in the future, it would already be in a convient place. Instead of in the WebP bindings. I can still move it there if you want

ManlyMarco: If it's not required for other formats right now I'd put it here and leave it even if future formats also need such a calculation.

ManlyMarco (approved):

PR #279, opened 2024-08-11
[ME] Fix loading of broken texture properties

RikkiBalboa:
It is apparently possible that ME saves a property with certain values as `null` that should never be `null`. I don't know how this would happen, but when this happens to a texture property, loading will fail when it hits this property.
The cause for the instance I was checking was because the method that converts normalmaps checks if the property is a normalmap property, but this fails if the property name is `null`.

`null` Values don't seem to be a problem for any other property types because they're just passed around until it hits the actual Unity method, which doesn't care about `null` values itself.

So to fix this I added some `null` checks to the texture property to prevent this.

### DIFF
--- a/src/MaterialEditor.Core/Core.MaterialEditor.CharaController.cs
+++ b/src/MaterialEditor.Core/Core.MaterialEditor.CharaController.cs
@@ -582,7 +582,7 @@ namespace KK_Plugins.MaterialEditor
                     for (var i = 0; i < properties.Count; i++)
                     {
                         var loadedProperty = properties[i];
-                        if (objectTypesToLoad.Contains(loadedProperty.ObjectType))
+                        if (objectTypesToLoad.Contains(loadedProperty.ObjectType) && !loadedProperty.NullCheck())
                         {
                             int? texID = null;
                             if (loadedProperty.TexID != null && importDictionary.TryGetValue((int)loadedProperty.TexID, out var importTextID))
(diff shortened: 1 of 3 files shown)


RikkiBalboa:
I don't know why the NormalMapManager shows so many changes, but there aren't any except for adding the try catch

ManlyMarco:
It's because of whitespace changes, try hiding whitespace on the github diff.

src/MaterialEditor.Core/Core.MaterialEditor.CharaController.cs
```diff
@@ -3127,7 +3127,7 @@ public MaterialTextureProperty(ObjectType objectType, int coordinateIndex, int s
             /// Check if the TexID, Offset, and Scale are all null. Safe to remove this data if true.
             /// </summary>
             /// <returns></returns>
-            public bool NullCheck() => TexID == null && Offset == null && Scale == null;
+            public bool NullCheck() => (TexID == null && Offset == null && Scale == null) || Property == null || MaterialName == null;
```
ManlyMarco, line 3130: Summary is no longer accurate and probably should be rewritten. Also ugly.

ManlyMarco (commented):
Seems good, only worry being that the NullCheck change might affect old code that relied on old behaviour, but it just needs testing in game.

ManlyMarco (approved):

PR #282, opened 2024-08-27
[Colliders] Reduce FindObjects

takahiro0327:
I reduced the number of calls to FindObjects.
I don't think that all FindObjects should be deleted. However, I fixed the FindObjects for Collider because the number of calls during scene loading was a little high.
Please review and merge.

Changes in loading time for huge scenes:

Before average 34.5
```
Scene loading completed: 34.8[s]
Scene loading completed: 34.3[s]
Scene loading completed: 34.6[s]
```

After average 33.62 -0.88
```
Scene loading completed: 33.8[s]
Scene loading completed: 33.9[s]
Scene loading completed: 33.5[s]
Scene loading completed: 33.6[s]
```

### DIFF
--- a/src/Colliders.Core/Core.Colliders.Controller.cs
+++ b/src/Colliders.Core/Core.Colliders.Controller.cs
@@ -33,10 +33,24 @@ namespace KK_Plugins
 
             private bool didSetStates;
 
+            private static HashSet<ColliderController> allControllers = new HashSet<ColliderController>();
+
             private float BreastSize => ChaControl.chaFile.custom.body.shapeValueBody[(int)BodyShapeIdx.BustSize];
             public bool BreastCollidersEnabled { get; set; }
             public bool SkirtCollidersEnabled { get; set; }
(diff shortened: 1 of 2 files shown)


src/Colliders.Core/Core.Colliders.Controller.cs
```diff
@@ -380,6 +395,16 @@ private void UpdateLegCollidersSkirtDB(DynamicBone dynamicBone)
...
             }
 #endif
+
+            /// <summary>
+            /// Get all active controllers
+            /// </summary>
+            static public IEnumerable<ColliderController> GetAllColliderControllers()
+            {
+                foreach (var controller in allControllers)
+                    if (controller && controller.enabled)
```
ManlyMarco, line 405: To match the behaviour of `FindObjectsOfType` it would need to be `activeInHierarchy` instead of `enabled`, since it only cares about the GameObject being active in hierarchy. If this is not an issue then it would be fine as it is however.

ManlyMarco (approved):

PR #286, opened 2024-09-11
[ME] Add AIT Keywords to the list

Hanmen-lab:
Added AIT Keywords to the list


ManlyMarco (commented):
Looks like this is fine to merge, but first shouldn't these properties also get added to `MaterialEditor.AI\Resources\default.xml`? It's almost identical to HS2 so I'd imagine they would work just as well in AI.

PR #288, opened 2024-09-28
[ShaderSwapper] Material name inclusion and exclusion rules based resolution

Petraller:
Added new feature to ShaderSwapper for swapping a shader to different shaders depending on the material name. Backward compatibility with old config file is maintained.

Example config file snippet:
```xml
<ShaderSwapper>
  <!-- Old structure for "Mapping" element, still supported -->
  <Mapping From="Shader Forge/main_skin" To="xukmi/SkinPlus">

  <!-- New structure for "Mapping" element, keep the old "From" attribute... -->
  <Mapping From="Shader Forge/main_skin">
    <!-- ...but introduce a "Rule" element to replace the "To" attribute -->
    <Rule Name="xukmi/SkinPlus" />
  </Mapping>

  <!-- Advanced usage, with includes and excludes -->
  <Mapping From="original_shader">
    <!-- Change to shader_0 if the material name is material_0 or material_1... -->
    <Rule Name="shader_0">
      <Include>
        <Entry>material_0</Entry>
        <Entry>material_1</Entry>
      </Include>
    </Rule>

    <!-- ...else, change to shader_1 if the material name is NOT material_2 or material_3... -->
    <Rule Name="shader_1">
      <Exclude>
        <Entry>material_2</Entry>
        <Entry>material_3</Entry>
      </Exclude>
    </Rule>

    <!-- ...else, change to shader_2 -->
    <Rule Name="shader_2" />
  </Mapping>
</ShaderSwapper>
```

### DIFF
--- a/src/ShaderSwapper.Core/ShaderSwapper.cs
+++ b/src/ShaderSwapper.Core/ShaderSwapper.cs
@@ -30,62 +30,190 @@ namespace KK_Plugins
         public const string PluginGUID = "com.deathweasel.bepinex.shaderswapper";
         public const string PluginName = "Shader Swapper";
         public const string PluginNameInternal = Constants.Prefix + "_ShaderSwapper";
-        public const string PluginVersion = "1.6";
+        public const string PluginVersion = "1.7";
         internal static new ManualLogSource Logger;
         private static ShaderSwapper Instance;
+
+        internal class SwapTargetList
+        {
+            internal class Rule
+            {
+                internal class EntryList
+                {
+                    public static string EntryElementName { get; } = "Entry";
+
+                    public List<string> Entries { get; set; } = new List<string>();
+
+                    public EntryList() { }
+
+                    public EntryList(IEnumerable<string> entries)
+                    {
+                        Entries = new List<string>(entries);
+                    }
+
+                    public EntryList(IEnumerable<XElement> collection)
+                    {
+                        Entries = new List<string>(collection.Select(e => e.Value));
+                    }
+
+                    public IEnumerable<XElement> ToElements()
+                    {
+                        return Entries.Select(e => new XElement(EntryElementName, e));
+                    }
+                }
+
+                public static string ElementName { get; } = "Rule";
+                public static string ShaderAttributeName { get; } = "Name";
+                public static string IncludeElementName { get; } = "Include";
+                public static string ExcludeElementName { get; } = "Exclude";
+
+                public string Shader { get; set; } = "MISSING SHADER NAME";
+                public EntryList Include { get; set; } = new EntryList();
+                public EntryList Exclude { get; set; } = new EntryList();
+
+                public Rule(string shader)
+                {
+                    Shader = shader;
+                }
+
+                public Rule(string shader, IEnumerable<string> include, IEnumerable<string> exclude)
+                {
+                    Shader = shader;
+                    Include = new EntryList(include);
+                    Exclude = new EntryList(exclude);
+                }
+
+                public Rule(XElement element)
+                {
+                    if (element == null || element.Name != ElementName)
+                        return;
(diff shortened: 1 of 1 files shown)


src/ShaderSwapper.Core/ShaderSwapper.cs
```diff
@@ -30,62 +30,190 @@ public class ShaderSwapper : BaseUnityPlugin
         public const string PluginGUID = "com.deathweasel.bepinex.shaderswapper";
         public const string PluginName = "Shader Swapper";
         public const string PluginNameInternal = Constants.Prefix + "_ShaderSwapper";
-        public const string PluginVersion = "1.6";
+        public const string PluginVersion = "1.7";
```
ManlyMarco, line 33: No version bumping in PRs please.
```suggestion
        public const string PluginVersion = "1.6";
```

src/ShaderSwapper.Core/ShaderSwapper.cs
```diff
@@ -30,62 +30,190 @@ public class ShaderSwapper : BaseUnityPlugin
...
         internal static new ManualLogSource Logger;
         private static ShaderSwapper Instance;
 
+        internal class SwapTargetList
+        {
+            internal class Rule
+            {
+                internal class EntryList
+                {
+                    public static string EntryElementName { get; } = "Entry";
```
ManlyMarco, line 43: How about naming it "Shader_Name" or similar instead in order to make it easier to understand on a glance? It would also allow for different elements like layer or GameObject name/path in the future. It could even be done with reflection I guess, e.g. "Material_xyz" which checks property "xyz" on the Material.

src/ShaderSwapper.Core/ShaderSwapper.cs
```diff
@@ -30,62 +30,190 @@ public class ShaderSwapper : BaseUnityPlugin
...
+                {
+                    var element = new XElement(ElementName, new XAttribute(ShaderAttributeName, Shader));
+                    if (Include.Entries.Count > 0)
+                        element.Add(new XElement(IncludeElementName, Include.ToElements()));
+                    if (Exclude.Entries.Count > 0)
+                        element.Add(new XElement(ExcludeElementName, Exclude.ToElements()));
+                    return element;
+                }
+
+                public bool TryGetToShader(Material material, out string toShader)
```
ManlyMarco, line 109: How about naming it "TryGetReplacementShader"? Or just add a summary to the methods to explain what they do.

ManlyMarco (commented):
Code looks good, maybe just needs some summary comments on a few methods.

The kinds of rules could be expanded a lot but it's fine if you don't want to do it for this PR.

ManlyMarco:
@Njaecha

src/ShaderSwapper.Core/ShaderSwapper.cs
Petraller: I think leaving the element name as "Entry" would be better, and instead changing the surrounding element name (i.e. "Include", "Exclude") to describe different behaviours and rules? Can definitely foresee this feature being expanded in the future. Thoughts?

Njaecha (approved):
_LGTM and back to BG3..._
Jokes aside: Tested and approved!
The configuration might be a bit complicated to understand for the common Koi user though. But I don't have any improvement ideas for that except maybe a tool that helps you write the config.

src/ShaderSwapper.Core/ShaderSwapper.cs
ManlyMarco: Whichever way works for you, really.

ManlyMarco (approved):

ManlyMarco:
Thanks for the PR! I guess the readme should be updated too now.

PR #292, opened 2024-12-16
[ME] Add manifest config `Hidden`; Add manifest template

acezen3d:
This is mainly to explicitly specify default values while keeping the properties hidden from the user.


src/MaterialEditor.Base/PluginBase.cs
```diff
@@ -290,15 +291,17 @@ public class ShaderPropertyData
...
             public float? MaxValue;
 
-            public ShaderPropertyData(string name, ShaderPropertyType type, string defaultValue = null, string defaultValueAB = null, string minValue = null, string maxValue = null)
+            public ShaderPropertyData(string name, ShaderPropertyType type, string defaultValue = null, string defaultValueAB = null, string hidden = null, string minValue = null, string maxValue = null)
             {
                 Name = name;
                 Type = type;
                 DefaultValue = defaultValue.IsNullOrEmpty() ? null : defaultValue;
                 DefaultValueAssetBundle = defaultValueAB.IsNullOrEmpty() ? null : defaultValueAB;
+                Hidden = bool.TryParse(hidden, out bool result);
```
ManlyMarco, line 304: This seems to be a mistake, the return value only shows if parsing was successful, not if the property is actually set to true.

ManlyMarco: Maybe something like
```suggestion
                Hidden = bool.TryParse(hidden, out bool result) && result;
```

mog1347: Yeah, this is a mistake.

acezen3d:
Tested on KK&KKS, good to go.

ManlyMarco (approved):

Issue #228, opened 2024-02-08
[MaterialEditor] Vestigial alpha_a and alpha_b sliders bug


Tarried:

Tarried:
Not entirely sure if this is intended:

Main_skin texture, if duplicated ('copy material'), will expose the alpha_a and alpha_b sliders on the dupicated copy, by default set to 0.

Adding an AlphaMask texture to the second material will fail to work as a result/the alphamask is ignored and no transparency is applied.

This is usually not an issue because you can just set those sliders to 0 to get the expected effect, but if you'd already changed to a different shader, those sliders are hidden but retain their 0 value, and so it isn't immediately obvious why AlphaMasking fails to do anything.

I read in the changelog that alpha_a and alpha_b sliders were removed a while back, so what would be the best way to solve this (obscure) issue? Or would trying to do so cause too many problems with clothing tops?

PR #295, opened 2024-12-28
[ME] Hide `AlphaMask` on all body materials, make `alpha_a` and `alpha_b` hidden everywhere

acezen3d:
This should probably solve #228.

### DIFF
--- a/src/MaterialEditor.Core/Core.MaterialEditor.cs
+++ b/src/MaterialEditor.Core/Core.MaterialEditor.cs
@@ -887,9 +887,7 @@ namespace KK_Plugins.MaterialEditor
 #if KK || EC || KKS
         public override bool CheckBlacklist(string materialName, string propertyName)
         {
-            if (materialName == "cf_m_body" || materialName == "cm_m_body")
-                if (propertyName == "alpha_a" || propertyName == "alpha_b" || propertyName == "AlphaMask")
-                    return true;
+            if (propertyName == "alpha_a" || propertyName == "alpha_b") return true;
             return false;
         }
 #endif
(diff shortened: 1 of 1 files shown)


acezen3d:
When copying the body material, the name of the new material becomes something like "cf_m_body.MECopy1". Such a name allows it to escape from previous blacklist rule and expose the properties in Material Editor.

RikkiBalboa (approved):

ManlyMarco (approved):

src/MaterialEditor.Core/Core.MaterialEditor.cs
```diff
@@ -886,10 +887,11 @@ private static Shader LoadShader(string shaderName, string assetBundlePath, stri
...
 #if KK || EC || KKS
         public override bool CheckBlacklist(string materialName, string propertyName)
-        {
-            if (materialName == "cf_m_body" || materialName == "cm_m_body")
-                if (propertyName == "alpha_a" || propertyName == "alpha_b" || propertyName == "AlphaMask")
-                    return true;
+        {
+            if (propertyName == "alpha_a" || propertyName == "alpha_b") return true;
+
+            if (Regex.IsMatch(materialName, @"^c[mf]_m_body(\.MECopy\d+)?$") && propertyName == "AlphaMask") return true;
```
ManlyMarco, line 893: Actually, can you use this constant just in case? https://github.com/IllusionMods/KK_Plugins/blob/d79e22073b31a9dac87964138386c2b2590309bd/src/MaterialEditor.Base/MaterialAPI.cs#L21

ManlyMarco (approved):

PR #298, opened 2025-01-10
[ME] Add tooltips

RikkiBalboa:
https://github.com/user-attachments/assets/c15508ae-dc70-4ec9-90b5-b4e26b468187

### DIFF
--- a/src/MaterialEditor.Base/UI/UI.Tooltip.cs
+++ b/src/MaterialEditor.Base/UI/UI.Tooltip.cs
@@ -1,0 +1,105 @@
+﻿using System;
+using System.Collections.Generic;
+using UILib;
+using UnityEngine;
+using UnityEngine.UI;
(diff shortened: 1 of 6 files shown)


src/MaterialEditor.Base/UI/UI.Tooltip.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (setActive)
+                SetActive(true);
+            LayoutRebuilder.ForceRebuildLayoutImmediate(panelTransform);
+        }
+
+        public void SetActive(bool active) 
+        {
+            if(MaterialEditorPluginBase.Showtooltips.Value)
+                Panel.gameObject.SetActive(active);
+        }
```
ManlyMarco, line 66: What if the setting is changed to false while a tooltip is being shown? Wouldn't it stay floating in space forever?

ManlyMarco (approved):

PR #302, opened 2025-01-23
[ME] Add `Category` attribute, `SortPropertiesByType`, `SortPropertiesByAlphabet` config and color work

acezen3d:
- Add property `Category` attribute to the shader manifest.
- Add two configs: `SortPropertiesByType` and `SortPropertiesByAlphabet` to control the property sorting.
- Add different colors to separator items to improve distinction in a lengthy context.
- Update shader manifest template.


acezen3d:
In order to avoid repeated grouping and sorting in UI drawing, a data holder is added.
In addition, for shaders that do not have property category info at all, we will no longer categorize them, meaning no "Uncategorized" label will be shown. This way, older shaders can remain unchanged.

ManlyMarco (approved):

PR #308, opened 2025-02-26
[ME] Hide categories when filtering properties and avoid empty properties

acezen3d:
The purpose of this change is to simplify the logic for hiding categories. Previously, a category would check whether it had any properties under it during filtering; if it did, the category would be displayed. The new logic simplifies this, categories will now be hidden whenever filtering is applied.

### DIFF
--- a/src/MaterialEditor.Base/UI/UI.cs
+++ b/src/MaterialEditor.Base/UI/UI.cs
@@ -467,7 +467,11 @@ namespace MaterialEditorAPI
                 filterList.RemoveAll(x => x.IsNullOrWhiteSpace());
 
                 filterListProperties = new List<string>(filterList);
-                filterListProperties = filterListProperties.Where(x => x.StartsWith("_")).Select(x => x.Trim('_')).ToList();
+                filterListProperties = filterListProperties
+                    .Where(x => x.StartsWith("_"))
+                    .Select(x => x.Trim('_'))
+                    .Where(x => !string.IsNullOrEmpty(x))
+                    .ToList();
                 filterList = filterList.Where(x => !x.StartsWith("_")).ToList();
             }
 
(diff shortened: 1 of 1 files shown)


src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -467,7 +467,11 @@ protected void PopulateList(GameObject go, object data, string filter = null)
                 filterList.RemoveAll(x => x.IsNullOrWhiteSpace());
 
                 filterListProperties = new List<string>(filterList);
-                filterListProperties = filterListProperties.Where(x => x.StartsWith("_")).Select(x => x.Trim('_')).ToList();
+                filterListProperties = filterListProperties
+                    .Where(x => x.StartsWith("_"))
+                    .Select(x => x.Trim('_'))
+                    .Where(x => !string.IsNullOrEmpty(x))
```
ManlyMarco, line 473: It's impossible for x to be empty because it has to start with _

src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -684,13 +688,11 @@ void PopulateListMaterial(Material mat, Projector projector = null)
 
                 foreach (var category in categories)
                 {
-                    // Blacklist and filter
-                    if (!category.Value.Any(p =>
-                        !Instance.CheckBlacklist(materialName, p.Name)
-                        && (filterListProperties.Count == 0 || filterListProperties.Any(fw => WildCardSearch(p.Name, fw))))
-                    ) continue;
```
ManlyMarco, line 691: Why remove the blacklist feature?

src/MaterialEditor.Base/UI/UI.cs
acezen3d: Filtering still activates when you enter only "_".

src/MaterialEditor.Base/UI/UI.cs
acezen3d: Sorry, I didn't provide the proper context. I've added a comment above. Completely hiding categories during filtering can help alleviate UI lag.

ManlyMarco (approved):

PR #311, opened 2025-03-23
TimelineFlowControl - Take deltaTime into account when jumping

Dave9031:
When cascading multiple "jumps" the playback would visually stop after the jump, since it only jumped to the commandTime which breaks continuity for loops. (even worse, sometimes it would happen while at other times not)
This commit fixes that and **makes jumps predictable**

### HOWEVER: This introduces a problem --> if there is a FlowCommand very close AFTER the jump for example at 00:00.00005 when the jump target is at 0:00 it MIGHT or might not get executed depending on where the playback was before the jump.
I am not sure how to handle it:
a) This is fine
b) Maybe do the jump in 0 steps? jump to commandTime, then "run through" all the interpolables between commandTime and commandTime + deltaTime
c) A much better and smarter solution :^)

image
image
image
image

out

### DIFF
--- a/src/TimelineFlowControl.Shared/FlowControlStateMachine.cs
+++ b/src/TimelineFlowControl.Shared/FlowControlStateMachine.cs
@@ -1,12 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using UnityEngine;
 
 namespace TimelineFlowControl
 {
     public class FlowControlStateMachine
     {
-        private const int _maxJumpDepth = 64;
+        private const int _maxJumpDepth = 128;
         private readonly Dictionary<string, float> _variableValues = new Dictionary<string, float>();
 
         private float _lastJumpTime;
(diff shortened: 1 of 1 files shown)


ManlyMarco (changes requested):
The fix is good but I'm worried about merging it because of the issue you've mentioned. It would make FlowCommands unpredictable in some cases (getting worse with lower FPS) and might break existing scenes. At the very least FlowCommands should be guaranteed to always be executed in the order they are on the timeline.

Dave9031:
@ManlyMarco 
I tried one approach that I could think of, briefly tested it with FPS locked to 4, 60, 137 aswell as using it with VideoExport
I added FlowCommands 0.005s after other FlowCommands

It indeed did not skip them like expected + frame timing when jumping seems to be precise so I am confident I can make this work

Please tell me if youre fine with this approach and I will iron it out and properly test it
And of course I welcome ideas and opinions! more eyes are always better
Thanks

src/TimelineFlowControl.Shared/FlowControlStateMachine.cs
```diff
@@ -6,32 +6,138 @@ namespace TimelineFlowControl
...
+                        }
+                    }
+                }
+                if (thereWasAJump) continue;
+                
+                
+                // TODO: detect if we loop over multiple times (when we have like 1 fps, which will make deltaTime huge)
+                float overflowedTargetTimeWithDeltaTime = targetTimeWithDeltaTime % Timeline.Timeline.duration;
+                
+                if (overflowedTargetTimeWithDeltaTime < targetTime) // Timeline will overflow/loop, so check keyframes starting from 0.0 aswell
```
Dave9031, line 108: Typo: should be "overflowedTargetTimeWithDeltaTime < targetTimeWithDeltaTime"

ManlyMarco:
That's pretty messy but it's probably as good as it gets without modifying Timeline to add new APIs ([which you can absolutely do)](https://github.com/IllusionMods/HSPlugins/). If you clean it up a bit more then I think it can be merged.

On an unrelated note, I was thinking of a way to make the jumps carry over state of interpolables from when the jump was called, so that you can jump to anywhere and the animation would transition smoothly. This approach might work for that but I think it would need modifications to Timeline itself.

PR #315, opened 2025-04-04
[ME] never load/set updateWhenOffscreen in KK

RikkiBalboa:
This property can crash KK when set (not sure if this also occurs on other games,, but it doesn't on KKS). This will prevent the following in KK:
- The saved property value from being loaded and thus applied on load
- Showing the UI element to set it
- The value from being set if some other plugin calls the method


ManlyMarco (approved):

Tarried:
Is there any way we can get this option back as a toggle from F1 menu? 

I've never had a crash I've attributed to using it, and it's has a very useful application for scene making where you would otherwise be fiddling with an object's origin location + KKPE to stop it from random disappearing in certain camera angles

ManlyMarco:
If you share scenes with that override there's a chance it may crash on other people's PCs.

PR #324, opened 2025-05-05
[ShaderSwapper] Add more shader presets

Splendide-Imaginarius:
This PR adds shader presets for Az, USS, and UTS shaders.

I've tested on KK and it seems to work fine. It builds without errors for KKS but I wasn't able to test there beyond that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -730,9 +730,9 @@ Boop the character by moving mouse over parts of their body, hair and clothes. U
 - [KK_ShaderSwapper]
 - [KKS_ShaderSwapper]
 
-Swap all shaders to the equivalent Vanilla Plus shader in the character maker or studio by pressing right ctrl + P.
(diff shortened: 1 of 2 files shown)


Njaecha:
I don't really understand why we want to add more hard-coded presets, when the point of the XML shenanigans was to create easily shareable configs, so that people could send others their SW setup.
https://github.com/IllusionMods/KK_Plugins/pull/324/commits/aa596785cc8e0c2303af737af1760055df5e3b73 look like upgrade though.

Njaecha (approved):
Doesn't really provide anything new except the Az compatibility for SetTesselationValue(Material mat), but also doesn't really impact anything. The average user won't probably notice a change, since they would have to manually change the path to one of the new XMLs in the settings.

LGTM

RikkiBalboa (approved):

PR #327, opened 2025-05-12
Use sdk project format

Keelhauled:
Still working on it. Can i delete the PostBuildEvents?


ManlyMarco:
Yes put the events in props as they should've always been.

You could probably put nuget packages into the props as well and switch them based on defines (KK EC etc).

Keelhauled:
I thought about that but realized it just makes things confusing and isnt even that useful since nuget packages are pretty much a one time deal. Same with project imports.

ManlyMarco (approved):

PR #332, opened 2025-05-29
[ME] Fix texture property timeline support on characters

### LINKED ISSUE #322: [KK] Materialediter f_m_hitomi_00 MainTex interpolables timeline save bug
When i add cf_m_hitomi_00 MainTex as interpolables in timeline, to change the eye overlay. It works fine when i hit play.
But when i save the scene, the eye overlay gets removed and presist no matter if i try to reload or restart studio.
I tried to update all my plugins, but the problem is still there.

My goal was to have the eye overlay change during animation.

Image
Image

RikkiBalboa:
By getting the referenced texture id's used by timeline directly, ME can more accurately keep textures that are still referenced. With the previous method textures on characters would still get purged on scene save.
The textures referenced by timeline are still saved in the scene itself (even when on a character) and will thus survive a character swap too (though they'll only work if they share renderers/materials)

fixes #322

### DIFF
--- a/src/MaterialEditor.Core.Studio/Core.MaterialEditor.TimelineCompatibilityHelper.cs
+++ b/src/MaterialEditor.Core.Studio/Core.MaterialEditor.TimelineCompatibilityHelper.cs
@@ -578,19 +578,32 @@ namespace MaterialEditorAPI
 
         private static void InitTimelineTypes()
         {
-            _timelineType = System.Type.GetType("Timeline.Timeline,Timeline", throwOnError: false);
-            if (_timelineType == null) return;
+            try
+            {
+                _timelineType = System.Type.GetType("Timeline.Timeline,Timeline", throwOnError: false);
+                if (_timelineType == null)
+                {
+                    isTimelineAvailable = false;
+                    return;
+                };
 
-            _interpolableType = System.Type.GetType("Timeline.Interpolable,Timeline", throwOnError: false);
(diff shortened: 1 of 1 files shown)


src/MaterialEditor.Core.Studio/Core.MaterialEditor.TimelineCompatibilityHelper.cs
```diff
@@ -534,5 +538,59 @@ internal static void RefreshInterpolablesList()
...
+            _interpolableType = System.Type.GetType("Timeline.Interpolable,Timeline", throwOnError: false);
+            _keyframeType = System.Type.GetType("Timeline.Keyframe,Timeline", throwOnError: false);
+
+            _GetAllInterpolables = _timelineType.GetMethod("GetAllInterpolables", AccessTools.all);
+
+            _interpolableOwner = _interpolableType.GetField("owner", AccessTools.all);
+            _interpolableId = _interpolableType.GetField("id", AccessTools.all);
+            _interpolableOci = _interpolableType.GetField("oci", AccessTools.all);
+            _interpolableKeyframes = _interpolableType.GetField("keyframes", AccessTools.all);
+            _keyframeValue = _keyframeType.GetField("value", AccessTools.all);
```
ManlyMarco, line 593: Add some sort of sanity checks in case any of these fields get changed in timeline.

ManlyMarco (approved):

PR #333, opened 2025-06-04
[Pushup] More outfits support for copying data between outfits

RikkiBalboa:


src/Pushup.Core/Core.Pushup.GUI.cs
```diff
@@ -202,16 +223,36 @@ private void RegisterCustomSubCategories(object sender, RegisterSubCategoriesEve
...
+            var coordinateList = Enum.GetNames(typeof(ChaFileDefine.CoordinateType)).ToList();
+
+            if (MoreOutfitsType != null)
+                ExtendList();
+
+            void ExtendList()
+            {
+                var moreOutfitsController = MoreOutfits.Plugin.GetController(MakerAPI.GetCharacterControl());
+                for (int coord = coordinateList.Count; coord < MakerAPI.GetCharacterControl().chaFile.coordinate.Length; coord++)
+                    coordinateList.Add(MoreOutfits.Plugin.GetController(MakerAPI.GetCharacterControl()).GetCoodinateName(coord));
```
ManlyMarco, line 248: Why not like this?
```suggestion
                var mcc = MakerAPI.GetCharacterControl();
                var moreOutfitsController = MoreOutfits.Plugin.GetController(mcc);
                for (int coord = coordinateList.Count; coord < mcc.chaFile.coordinate.Length; coord++)
                    coordinateList.Add(moreOutfitsController.GetCoodinateName(coord));
```

src/Pushup.Core/Core.Pushup.cs
```diff
@@ -24,6 +25,7 @@ public partial class Pushup : BaseUnityPlugin
         public const string Version = "1.4.0";
         internal static new ManualLogSource Logger;
 
+        private static Type MoreOutfitsType;
```
ManlyMarco, line 28: I would add a soft BepInDependency too just to be clear that there is a non-trivial interaction.

PR #343, opened 2025-07-05
[ME] Add `AnisoLevel`, `FilterMode`, `WrapMode` config for texture properties

acezen3d:
As the title said.

### DIFF
--- a/src/MaterialEditor.Base/NormalMapManager.cs
+++ b/src/MaterialEditor.Base/NormalMapManager.cs
@@ -159,6 +159,9 @@ namespace MaterialEditorAPI
                     Graphics.Blit(readableTex, rt);
                     RenderTexture.active = cur;
 
+                    rt.wrapMode = src.wrapMode;
+                    rt.anisoLevel = src.anisoLevel;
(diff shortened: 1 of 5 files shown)


src/MaterialEditor.Base/NormalMapManager.cs
```diff
@@ -42,7 +42,7 @@ public bool ConvertNormalMap(ref Texture tex, string propertyName, bool unpack)
         {
             try
             {
-                if (!NormalMapProperties.Any(x => propertyName.Contains(x)))
+                if (!NormalMapProperties.Contains(propertyName))
```
acezen3d, line 45: Like in these part:

https://github.com/IllusionMods/KK_Plugins/blob/02b496b7df8e1e58c60bbbcb45c60b4790365730/src/MaterialEditor.Core/Core.MaterialEditor.cs#L989

https://github.com/IllusionMods/KK_Plugins/blob/02b496b7df8e1e58c60bbbcb45c60b4790365730/src/MaterialEditor.Core/Core.MaterialEditor.cs#L999

ManlyMarco (approved):

PR #350, opened 2025-07-22
[ME] Don't add category headers when the category is empty

RikkiBalboa:


src/MaterialEditor.Base/UI/UI.cs
```diff
@@ -678,26 +678,26 @@ void PopulateListMaterial(Material mat, Projector projector = null)
 
                 foreach (var category in categories)
                 {
+                    var properties = category.Value.Where(x => mat.HasProperty($"_{x.Name}"));
                     if (
                         filterListProperties.Count == 0
                         && (categories.Count > 1 || category.Key != PropertyOrganizer.UncategorizedName)
+                        && properties.Count() > 0
```
ManlyMarco, line 685: More faster
```suggestion
                        && properties.Any()

```

ManlyMarco (approved):

PR #354, opened 2025-07-30
Add SpawnLocker (code provided by BitMagnet)

ManlyMarco:
Includes original code that was refactored to fit the style used in this repo.
~~Needs regression testing with the original plugin.~~ All good.


src/SpawnLocker.Core/SpawnLockData.cs
```diff
@@ -0,0 +1,53 @@
+using ExtensibleSaveFormat;
```
acezen3d, line 1: Shouldn't this work?

acezen3d (commented):
So this must be submitted?

PR #361, opened 2025-09-15
[ME] Fix KeywordPropertyList duplication in scenes

krypto5863:
Ayup


src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
```diff
@@ -357,6 +358,18 @@ protected override void OnSceneLoad(SceneOperationKind operation, ReadOnlyDictio
                     }
                 }
             }
+
+            var fixedKeywordList = MaterialKeywordPropertyList
```
ManlyMarco, line 362:
```suggestion
            // Clean up scenes saved in buggy versions of ME that duplicated keyword props on 
            // scene loads, causing massive file sizes. `.First()` should always keep the latest user edit.
            var fixedKeywordList = MaterialKeywordPropertyList
```

ManlyMarco (approved):

PR #364, opened 2025-09-17
Fixed group by filtering, moved duplicate removal to it's own method a…

krypto5863:
…nd conditionalized it by data version so it's only run on older data.


ManlyMarco (approved):
Good idea with using data version

PR #370, opened 2025-10-30
[ME] Local / deduped textures

starstormhun:


src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
```diff
@@ -46,11 +48,46 @@ public class SceneController : SceneCustomFunctionController
...
+                    wasLoading = isLoading;
+                }
+                else if (counter > 0 && !isLoading)
+                {
+                    --counter;
+                    if (counter <= 0) DedupedTextureData = null;
+                }
+                yield return null;
+            }
+        }
```
ManlyMarco: Why not use the SceneLoad event instead? You can then invoke delayed from the event handler, if it's necessary.

src/Shared.TextureContainer/Shared.TextureContainer.cs
```diff
@@ -9,7 +9,7 @@ namespace KK_Plugins
     /// </summary>
     public sealed class TextureContainer : IDisposable
     {
-        private TextureContainerManager.Token _token;
+        internal TextureContainerManager.Token _token;
```
ManlyMarco: This is naughty. If you need the hash then expose just the hash in a new method or prop, eg GetTextureHash.

src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
```diff
@@ -30,13 +32,13 @@ public class SceneController : SceneCustomFunctionController
         private readonly List<MaterialFloatProperty> MaterialFloatPropertyList = new List<MaterialFloatProperty>();
         private readonly List<MaterialKeywordProperty> MaterialKeywordPropertyList = new List<MaterialKeywordProperty>();
         private readonly List<MaterialColorProperty> MaterialColorPropertyList = new List<MaterialColorProperty>();
-        private readonly List<MaterialTextureProperty> MaterialTexturePropertyList = new List<MaterialTextureProperty>();
+        internal readonly List<MaterialTextureProperty> MaterialTexturePropertyList = new List<MaterialTextureProperty>();
```
ManlyMarco: Not a fan of exposing internals specific to scenes, ideally there would be a cleaner way to do it.

src/MaterialEditor.Core/Core.MaterialEditor.cs
```diff
@@ -1019,6 +1171,449 @@ internal static void ConvertNormalMap(Material material, string propertyName)
...
+                    if (at == pattern.Length) return foundPosition;
+                } else
+                {
+                    at = 0;
+                }
+            }
+            return -1;
+        }
+
+        internal static IEnumerator AuditLocalFilesDone()
```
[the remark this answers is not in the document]
starstormhun: Fair point

src/MaterialEditor.Core/Core.MaterialEditor.cs
```diff
@@ -165,6 +279,31 @@ public override void Awake()
...
 #endif
             RendererCachingEnabled = Config.Bind("Config", "Renderer Cache", true, "Turning this off will fix cache related issues but may have a negative impact on performance.");
+
+            // Texture saving configs
+            ConfigLocalTexturePath = Config.Bind("Textures", "Local Texture Path Override", "", new ConfigDescription($"Local textures will be exported to / imported from this folder. If empty, defaults to {LocalTexturePathDefault}.\nWARNING: If you change this, make sure to move all files to the new path!", null, new ConfigurationManagerAttributes { Order = 10, IsAdvanced = true }));
+            ConfigLocalTexturePath.SettingChanged += ConfigLocalTexturePath_SettingChanged;
+
+            CharaLocalTextures.Activate();
+            TextureSaveTypeChara = Config.Bind("Textures", "Chara Texture Save Type", CharaLocalTextures.SaveType, new ConfigDescription("Texture save type for characters set in the Modding API.", null, new ConfigurationManagerAttributes { Order = 7, IsAdvanced = true, CustomDrawer = new Action<ConfigEntryBase>(KKAPISettingDrawer<CharaTextureSaveType>), HideDefaultButton = true }));
```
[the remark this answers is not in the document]
starstormhun: It's only for the autosaves, not the normal saves.
...I guess the autosave overrides can be added to KKAPI instead, then.

src/MaterialEditor.Core.Studio/Core.MaterialEditor.SceneController.cs
starstormhun: Because it was 4AM probably and I didn't wanna test which order Scene / Characontrollers fired their loads... :D
But it seems charas always go first, which is perfect, so I can just clear the variable when scenecontroller is done

src/MaterialEditor.Core/Core.MaterialEditor.cs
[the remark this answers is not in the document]
ManlyMarco: Then it should explicitly state that it's for Autosave

ManlyMarco (approved):

ManlyMarco (approved):

PR #387, opened 2026-05-03
ModBoneImplantor support for AccessoryClothes

### LINKED ISSUE #5: not working with AccessoryClothes
current versions of KK ModBoneImplantor and KK AccessoryClothes (not sure if any other plugins affect this), item made with KK Modding Tools

Tried to get it working on an AccessoryClothes accessory, the bone implant script seems to be ignored.  No message in the output log.  The part weighted to the body bones works fine but the part weighted to the implant bones is stretched to the floor.  (it works fine when I remake it as a clothes item and it tells me it implanted 7 bones)

Kokaiinum:
What it says on the tin. Only KK for now because I'm not certain what other games need the patch. KKS doesn't seem to, I'm not sure about EC/HS2/AI. Should fix https://github.com/IllusionMods/ModBoneImplantor/issues/5


src/AccessoryClothes.KK/Hooks.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace KK_Plugins
+{
+    internal class Hooks
+    {
+        internal static void InstallHooks() 
+        {
+            Harmony harmony = new Harmony("AccessoryClothesHarmony");
+
+            Type implantorType = Traverse.Create(AccessTools.TypeByName("ModBoneImplantor.ModBoneImplantor"))
+                .Type("AssignedAnotherWeightsHooks").GetValue<Type>();
```
ManlyMarco: This can be simplified in a couple different ways.

This avoids missing type warning in log if ModBoneImplantor is not installed:
```suggestion
            Type implantorType = Type.GetType("ModBoneImplantor.ModBoneImplantor, ModBoneImplantor", false)?.GetNestedType("AssignedAnotherWeightsHooks", BindingFlags.Public | BindingFlags.NonPublic);
```

This keeps the warnings while being more clear what type is got:
```suggestion
            Type implantorType = AccessTools.TypeByName("ModBoneImplantor.ModBoneImplantor+AssignedAnotherWeightsHooks");
```

ManlyMarco (commented):
Looks good but it would benefit from some code comments to explain what is happening and why.

Kokaiinum:
I opted for your first suggestion, better to avoid potentially confusing the user, I think. If they need MBI it'll be pretty obvious anyway (and they'll probably have it unless their install is from last decade).

However, I need to admit, that after some additional testing I think this fix should be extended to KKS and EC as well. The same fix should work My previous testing on KKS had some errors in methodology, but the problem does in fact exist there.

ManlyMarco (approved):

PR #389, opened 2026-05-06
Add material editor button for eye settings

Mixiel1:
Currently there is no easy way to edit eye settings, one has to go via Body -> Material Editor (All) -> type "eye".

And since I'm doing this constantly as many eye settings are missing from the default view, I want to add a button to the Left/Right Eye categories that directly opens the Material Editor with the "eye" filter applied.

### DIFF
--- a/KK_Plugins.sln
+++ b/KK_Plugins.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.8.34309.116
+# Visual Studio Version 18
+VisualStudioVersion = 18.5.11723.231 stable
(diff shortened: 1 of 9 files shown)


KK_Plugins.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.8.34309.116
+# Visual Studio Version 18
+VisualStudioVersion = 18.5.11723.231
```
ManlyMarco: Unnecessary change.
```suggestion
# Visual Studio Version 17
VisualStudioVersion = 17.8.34309.116
```

ManlyMarco (approved):
No idea what's going on with your commit history but the change is good. Thanks for the PR!

PR #392, opened 2026-05-31
Enhanced MaterialEditor UI (HS2/KKS)

KumihoArts:
Opt-in UI overhaul for the MaterialEditor window. Everything new is behind config options, so if you don't turn anything on it behaves like vanilla. It's pretty much all in MaterialEditor.Base plus a one line change in Core.MaterialEditor.Studio.cs.

Built and tested in HS2, ported to KKS. KK/AI/PH/EC compile but aren't tested.

What's in it:
- Collapsible renderer and material sections, and the collapse state sticks around
- Sticky section header bars while you scroll
- Left click a collapse button folds that one row, right click folds the whole tree
- Texture preview side panel: dockable so it follows the window, resizable, open/close, remembers its position
- Dark mode, a hacker theme, window and tooltip opacity sliders, custom UI fonts
- Drag to resize the window from a grip in the bottom right corner, with a sane minimum size
- Inline renderer enable toggle
- Search filter with a persist between objects toggle. Typing a property name without the underscore now searches properties instead of showing nothing
- Edited properties show in bold/accent text, and reset buttons light up when there's something to reset
- Copy a material's edits and paste onto one material with left click Paste, or pick several targets from a list with right click Paste
- Every material row shows which renderer(s) use it, and every renderer row lists its materials in the tooltip

Updated to have a go at the earlier review feedback (resize, collapse behaviour, filter, material/renderer clarity, paste target).


tomTom1010EEE:
Can you make it have the similar function as Liltoon does such as folding a specific section (catagory) and default folding for seldomly used feature and default presets for a shader (written by shader developer in manifest). 
I am developing a Liltoon recode for KKS, and was astonished with so many settable params that we haven't seen in unity editor cause they are folded by default.

ManlyMarco:
Note: Everyone can get a build of this PR from the actions - https://github.com/IllusionMods/KK_Plugins/actions/runs/26717752417

- The UI needs a way to be resized by mouse drag, it's too small by default.
- The collapse buttons at the bottom seem redundant other than collapse all. The - button next to each item can be clicked to collapse just that item already, and it should be possible to right click it to collapse the whole tree.
- It's unclear what material is under what renderer and vice-versa.
- You can copy a material but not choose what renderer to paste it on.
- There is still no visual indication of what a renderer/material is actually controlling. Something like what https://github.com/ManlyMarco/IL_SliderHighlight does as you hover over a material/renderer name would work. This is probably out of scope for this PR though.
- The filter box is still kind of ass, filtering for properties looks to be broken
- No property names?
image

PR #401, opened 2026-08-08
[ME] Add opt-in manifest schema 2 UI metadata

NoxOwlZzz:
## Description

- Add opt-in `<MaterialEditor SchemaVersion="2">` metadata for `DisplayName`, `UiLevel`, and `ShowIf`.
- Add Float-backed `Boolean` and `Enum` controls using the reviewed canonical syntax: optional `Invert="true"` for Boolean and `Enums="Label,Value,..."` for Enum.
- Make the Basic/Advanced selector independent per shader and identify visible Advanced property rows with a vertical blue accent bar.
- Preserve declaration-driven property/category ordering instead of adding manifest-only order fields.
- Add focused documentation, an updated schema 2 example manifest, and regression tests.

## Motivation and Context

Shader authors currently have to rely on internal property names and legacy controls. This adds an explicitly versioned, backward-compatible UI metadata layer so manifests can provide clearer labels, conditional visibility, and simple Float-backed controls without changing save data or material-edit paths.

Sideloader's outer `<manifest schema-ver="1">` remains unchanged. Missing, invalid, or unsupported Material Editor `SchemaVersion` values retain schema 1 behavior. Shader-specific schema 2 metadata is prevented from leaking into the global legacy fallback.

## Review Feedback Addressed

- Removed `Dropdown` as a schema 2 type alias; use `Type="Enum"`.
- Removed manifest-only `Order` and `CategoryOrder`; declaration order remains authoritative.
- Replaced nested options with Unity-style `Enums="Label,Value,..."` pairs and invariant validation.
- Use `Boolean` as the sole Float-backed Boolean data type; manifests use `Type="Boolean"`.
- Boolean writes only `0` or `1` after an explicit checkbox change; optional `Invert="true"` only reverses the displayed state. Existing values are not rewritten when the UI opens or refreshes.
- Enum values are matched by direct equality only. Unmatched values are displayed without being changed, and an explicit choice writes exactly its declared value, including when resolving a mixed selection.
- Scoped Basic/Advanced state and its selector per shader, with a vertical accent bar.
- Deferred Cubemap support to a separate follow-up PR, as requested.
- Relaxed two formatting-sensitive source-contract assertions raised in the latest review.

## How Has This Been Tested?

- Material Editor metadata regression suite: PASS.
- Release build of `MaterialEditor.API`: PASS, including its normal PublicApiAnalyzers configuration.
- Release builds for AI, EC, HS2, KK, KKS, and PH: PASS.
- Schema 2 example manifest parsed as valid XML.
- `git diff --check origin/master...HEAD`: PASS.

Existing XML-documentation warnings remain unchanged; no build errors were introduced.

## Screenshots (if appropriate)

Not included; this PR is focused on the parser/UI contract and manifest authoring workflow.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Compatibility

- No shipped public API or save-data format changes.
- Boolean and Enum edits continue through existing Float paths.
- Schema 1 manifests preserve their existing behavior.
- The reviewed schema 2 syntax replaces only unshipped PR syntax.

## Non-goals

- Cubemap support; it will be proposed in a separate follow-up PR.
- No Color/Vector distinction in this PR.
- No broader local UI redesign, theme, sidebar, renderer-collapse, or deployment changes.
- No persisted Basic/Advanced preference.
- No changes to legacy `DefaultValue` application or material serialization semantics.

### DIFF
--- a/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
+++ b/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
@@ -22,7 +22,7 @@ Schema 2 adds optional attributes to `<Property>`:
 - `DisplayName`: label shown in Material Editor; defaults to `Name`.
 - `UiLevel`: `Basic` (the default) or `Advanced`. When a shader declares at
   least one Advanced property, its shader row shows a Basic/Advanced selector.
-  Advanced property labels are prefixed with `[A]` when visible. The selection
+  Advanced property rows show a vertical accent bar when visible. The selection
   is maintained independently for each shader during the current session.
 
 ## Float-backed controls
--- a/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
+++ b/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
@@ -20,28 +20,31 @@ example described here.
 Schema 2 adds optional attributes to `<Property>`:
 
 - `DisplayName`: label shown in Material Editor; defaults to `Name`.
-- `Order`: property order within its category. Explicit values appear first,
-  from lowest to highest; ties keep declaration order.
-- `CategoryOrder`: category order derived from its properties. Explicit values
-  appear first, from lowest to highest. Use the same value on every property in
-  a category; if values disagree, the first explicit value declared wins.
-- `UiLevel`: `Basic` (the default) or `Advanced`.
+- `UiLevel`: `Basic` (the default) or `Advanced`. When a shader declares at
+  least one Advanced property, its shader row shows a Basic/Advanced selector.
+  Advanced property labels are prefixed with `[A]` when visible. The selection
+  is maintained independently for each shader during the current session.
 
 ## Float-backed controls
 
-`Type="Toggle"` is an alias for a Float property using the toggle editor.
-`OffValue` and `OnValue` select the float values written for each state and
-default to `0` and `1`.
+`Type="Toggle"` is an alias for a Float property using the toggle editor. It
+writes `0` for off and `1` for on. Set `Invert="true"` to reverse the displayed
+toggle while keeping the stored values limited to `0` and `1`.
 
-`Type="Enum"` and `Type="Dropdown"` are equivalent aliases for a Float
-property using the enum editor. Declare choices as direct children:
+`Type="Enum"` is an alias for a Float property using the enum editor. Declare
+choices as alternating labels and numeric values in the `Enums` attribute,
+matching Unity's enum-property style:
 
 ```xml
-<Property Name="BlendMode" Type="Enum" DisplayName="Blend mode">
-  <Option Value="0" DisplayName="Opaque"/>
-  <Option Value="1" DisplayName="Cutout"/>
-</Property>
+<Property Name="BlendMode"
+          Type="Enum"
+          DisplayName="Blend mode"
+          Enums="Opaque,0,Cutout,1"/>
 ```
+
+Labels and values must appear in pairs. Labels cannot be empty, values must be
+finite invariant-culture numbers, and duplicate labels or values are ignored
+with a warning.
 
 The aliases are case-insensitive, but only recognized when
 `SchemaVersion="2"` is active.
--- a/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
+++ b/Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
@@ -27,9 +27,11 @@ Schema 2 adds optional attributes to `<Property>`:
 
 ## Float-backed controls
 
-`Type="Toggle"` is an alias for a Float property using the toggle editor. It
-writes `0` for off and `1` for on. Set `Invert="true"` to reverse the displayed
-toggle while keeping the stored values limited to `0` and `1`.
+`Type="Boolean"` is a Float-backed Boolean control. It writes `0` for off and
+`1` for on. The optional `Invert="true"` attribute reverses the displayed state
+while keeping stored values limited to `0` and `1`; it defaults to `false`.
+`Type="Toggle"` remains accepted as a read-compatible alias, but new manifests
+should use `Boolean`.
 
 `Type="Enum"` is an alias for a Float property using the enum editor. Declare
 choices as alternating labels and numeric values in the `Enums` attribute,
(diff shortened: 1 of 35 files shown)


Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
```diff
@@ -0,0 +1,72 @@
...
+  a category; if values disagree, the first explicit value declared wins.
+- `UiLevel`: `Basic` (the default) or `Advanced`.
+
+## Float-backed controls
+
+`Type="Toggle"` is an alias for a Float property using the toggle editor.
+`OffValue` and `OnValue` select the float values written for each state and
+default to `0` and `1`.
+
+`Type="Enum"` and `Type="Dropdown"` are equivalent aliases for a Float
```
acezen3d: Dropdown is not a data type, compared with other data types such as Float, Texture, Color, etc. I suggest removing this alias.

Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
```diff
@@ -0,0 +1,72 @@
...
+example described here.
+
+## Property metadata
+
+Schema 2 adds optional attributes to `<Property>`:
+
+- `DisplayName`: label shown in Material Editor; defaults to `Name`.
+- `Order`: property order within its category. Explicit values appear first,
+  from lowest to highest; ties keep declaration order.
+- `CategoryOrder`: category order derived from its properties. Explicit values
```
acezen3d: Adding a `Category` field directly to the `Property` tag, rather than defining it separately or using a nested structure (category->property), was a reasonable choice for maintaining compatibility with the legacy manifest format without breaking it. However, adding `CategoryOrder` to the `Property` tag as well feels rather odd.

Moreover, both the property order and the category order can currently be controlled by shader authors simply by arranging the manifest entries in the desired order. So these two fields don't provide much value. I suggest removing them.

Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
```diff
@@ -0,0 +1,72 @@
...
+
+`Type="Enum"` and `Type="Dropdown"` are equivalent aliases for a Float
+property using the enum editor. Declare choices as direct children:
+
+```xml
+<Property Name="BlendMode" Type="Enum" DisplayName="Blend mode">
+  <Option Value="0" DisplayName="Opaque"/>
+  <Option Value="1" DisplayName="Cutout"/>
+</Property>
+```
```
acezen3d, line 43: I suggest changing the way type Enum is defined in the manifest and directly adopting the same enum definition style used by Unity (https://docs.unity3d.com/cn/2022.1/ScriptReference/MaterialPropertyDrawer.html), as shown below. The current nested structure is not friendly for shaders with a large number of enums. With the proposed format, shader authors can simply copy and paste the definitions from shader code.
This also avoids having to decide whether the name of an `Option` should be `Name` or `DisplayName`.

```xml
<Property Name="BlendMode" Type="Enum" DisplayName="Blend mode" Enums="Opaque,0,Cutout,1"/>
```

Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
```diff
@@ -0,0 +1,72 @@
...
+- `CategoryOrder`: category order derived from its properties. Explicit values
+  appear first, from lowest to highest. Use the same value on every property in
+  a category; if values disagree, the first explicit value declared wins.
+- `UiLevel`: `Basic` (the default) or `Advanced`.
+
+## Float-backed controls
+
+`Type="Toggle"` is an alias for a Float property using the toggle editor.
+`OffValue` and `OnValue` select the float values written for each state and
+default to `0` and `1`.
```
acezen3d: Toggle corresponds to a boolean, which logically only has two states: 0 (off) and 1 (on). There shouldn't be any logic for defining the on value and the off value; if that is needed, then the Enum type should be used instead.
If a Toggle needs to be inverted, i.e. on corresponds to 0 and off corresponds to 1, I suggest either adding the ToggleOff type to stay consistent with Unity, or adding an `Invert` field.

acezen3d (commented):
Suggestions for the fields added to the `Property` tag.

Guides/Material Editor Guide/MANIFEST_SCHEMA_V2.md
```diff
@@ -0,0 +1,72 @@
...
+
+Schema 2 adds optional attributes to `<Property>`:
+
+- `DisplayName`: label shown in Material Editor; defaults to `Name`.
+- `Order`: property order within its category. Explicit values appear first,
+  from lowest to highest; ties keep declaration order.
+- `CategoryOrder`: category order derived from its properties. Explicit values
+  appear first, from lowest to highest. Use the same value on every property in
+  a category; if values disagree, the first explicit value declared wins.
+- `UiLevel`: `Basic` (the default) or `Advanced`.
```
acezen3d: Discussed changes：
- Add some visual indicator to Advanced properties.
- Make the Basic/Advanced switch per-shader instead of global.

acezen3d (commented):
Discussed changes to `UiLevel` and the Basic/Advanced switch.

acezen3d (approved):
Thanks for making all the changes. I really appreciate it.
Oh, one more thing: Toggle should be renamed to Boolean to keep it consistent with the other data types. This gives us:
Texture, Color, Float, Boolean, Enum, Keyword, Cube/Cubemap (if cubemap support is added later)

Also, I've been wondering whether it's necessary to distinguish between Color and Vector.

ManlyMarco:
I will merge once all remaining comments are addressed.

NoxOwlZzz:
Addressed the remaining Boolean/Enum behavior: Enum matching now uses direct equality only, unmatched values remain untouched during UI open/refresh, explicit Enum selections write the exact declared value, and Boolean writes only 0 or 1 after a user checkbox change. Cubemap support remains deferred to a separate follow-up PR. Tests and Release builds pass.

NoxOwlZzz:
Removed the unreleased `Toggle` manifest spelling entirely. `Type="Boolean"` is now the only supported Boolean manifest type, and the parser, documentation, and regression tests have been updated accordingly. Tests and Release builds pass.

acezen3d (commented):
Looks good to me.

ManlyMarco (approved):
Great work, thank you for the PR.

PR #402, opened 2026-08-12
[ME] Integrate the unified Material Editor prototype

NoxOwlZzz:
## Summary

- Integrates the complete Material Editor UI prototype: responsive three-panel layout, pooled and virtualized rows, category navigation, independent renderer/material/category/subcategory folding, search, clearer controls and tooltips, and the accumulated UI bug fixes.
- Simplifies manifest presentation by removing Basic/Advanced, EnableIf, and SwitchProperty. Category and Subcategory remain optional and presentation-only, ShowIf stays explicit, and legacy manifests keep their existing behavior.
- Completes Boolean, Enum, Vector, and Cubemap handling across the API, UI, clipboard, reset, and Maker/Studio persistence.
- Hardens copy/paste and imports with type checks, real completion results, transactional rollback, stable original-value remapping, bounded cache/lease ownership, and incremental processing.
- Adds Radiance RGBE .hdr Cubemap import through RGBAHalf while preserving PNG import. Export remains PNG/SDR.
- Includes performance/lifetime improvements, regression coverage, and updated documentation.

## Validation

- API and KK/PH/EC/KKS/AI/HS2 Release builds
- MetadataTests and PerformanceHarness


src/MaterialEditor.Base/CopyContainer.cs
```diff
@@ -151,6 +318,10 @@ public class MaterialTextureProperty
             /// </summary>
             public byte[] Data;
             /// <summary>
+            /// Kind of texture represented by Data.
+            /// </summary>
+            internal MaterialAPI.ShaderPropertyType TextureKind;
```
acezen3d: If we treat Cubemap as a type distinct from Texture (Texture2D), why are we reusing some Texture type code logic in some places but not in others?
https://github.com/IllusionMods/KK_Plugins/pull/402/changes#diff-c94cbc50352794afb2114bb2caa5289691ead53bae89af831c31df0a71d10875R106
https://github.com/IllusionMods/KK_Plugins/pull/402/changes#diff-c94cbc50352794afb2114bb2caa5289691ead53bae89af831c31df0a71d10875R140
If we follow the code here, the manifest should be changed to `<Property type="Texture" kind="Cubemap/Cube"/>`

What if we need to handle Texture3D in the future?
The Enum and Boolean types added earlier can, in a sense, be justified by the fact that they are fundamentally both Floats, so we added a more complex layer of abstraction around them (although they could still be handled separately, which would actually be simpler). 

A decision needs to be made here: either handle Cubemap separately so that they remain completely independent of Texture, or treat them as a subtype of Texture, with manifest supporting the two corresponding forms:
`<Property type="Cubemap"/>`
`<Property type="Texture"/>`
`<Property type="Texture3D"/>`
or
`<Property type="Texture" kind="2D (default value)">`
`<Property type="Texture" kind="Cube"/>`
`<Property type="Texture" kind="3D"/>`

Then the code should then be made consistent with whichever approach we choose.

acezen3d (commented):
Thoughts on the code and API design.

src/MaterialEditor.Base/CopyContainer.cs
NoxOwlZzz: I separated `Texture`, `Cubemap`, and future `Texture3D` types.

Commit `87a70049` gives Cubemap its own model, persistence, UI, operations, and original-state tracking, while keeping the existing Texture2D behavior and API unchanged.

Only generic storage/conversion helpers are shared. `Texture3D` is not implemented yet. All metadata/API checks and six Material Editor Release targets pass.

tomTom1010EEE:
This implementationintroduces several architectural coupling points and layout assumptions that may make future changes harder and may cause regressions at narrow window widths or in multi-material workflows.

The most urgent areas are:

1. Responsive layout for the new controls.
2. Correct `ShowIf` behavior and refresh scope.
3. Separation of manifest-domain logic from the UI layer.
4. Consolidation of built-in and extension property-editor infrastructure.

## Findings

### 1. Manifest-domain logic is located in and depended on through the UI layer

**Severity: High**

`UI.MaterialPropertyMetadata.cs` currently contains several different concepts:

- Manifest metadata models.
- Schema parsing and validation.
- Condition parsing and evaluation.
- Enum value models.
- Float-backed value conversion policies.

These types are then used by `PluginBase.cs` and `Core.MaterialEditor.Shaders.cs`. This creates an inverted dependency in which core manifest-loading code depends on a file identified as part of the UI layer.

This will make non-UI manifest validation, alternative frontends, and future schema evolution harder. A clearer structure would be:

```text
Manifest/
  ManifestPropertyDefinition.cs
  ManifestMetadataParser.cs
  ManifestCondition.cs
  ManifestValidation.cs
  ShaderPropertyFallbackPolicy.cs

Presentation/
  PropertyVisibilityPolicy.cs
  ShaderUiModeState.cs

UI/
  Views, controls, binders, and layout only
```

Relevant implementation: [`UI.MaterialPropertyMetadata.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.MaterialPropertyMetadata.cs)

### 2. Adding a new editor type requires changes across too many files

**Severity: High**

Adding the new Enum and float-backed Boolean editors required coordinated changes to the parser, property data, descriptors, row models, row-model factory, control set, view factory, and binder.

This is a shotgun-surgery pattern: every future editor type, such as Vector, HDR Color, Cubemap, or Flags Enum, will require changes across many unrelated files.

There is also duplicated identity infrastructure. `MaterialEditorPropertyEditorIds.Boolean` already exists, while PR #401 adds `ShaderPropertyEditorIds.Boolean` with the same string value. Their consistency currently depends on manually keeping duplicated constants synchronized.

The built-in manifest editors also use hard-coded branching while extension properties use `MaterialEditorExtensionRegistry`, resulting in two parallel editor architectures.

A unified strategy registry would reduce this coupling:

```text
EditorId
  -> IPropertyEditorStrategy
       - ValidateMetadata
       - CreateModel
       - BindControls
       - ResolveStoredValue
```

Relevant implementation: [`UI.PropertyDescriptor.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.PropertyDescriptor.cs#L117)

### 3. `ShowIf` changes trigger a full Material Editor presentation rebuild

**Severity: High**

When a controlling property changes, the implementation defers for one frame and then calls the full `PopulateList()` path. This rebuilds all material, shader, category, and property rows rather than updating only the affected conditional rows.

Potential consequences include:

- The visible property at the same scroll position can change when rows are inserted or removed above it.
- Keyboard focus and selected controls can be lost.
- An open dropdown can be destroyed and rebound.
- Categories and row mappings are repeatedly recalculated.
- Large selections can incur unnecessary rebuild cost.

At minimum, the UI should preserve and restore a stable property/category anchor rather than only retaining the absolute scroll offset. Longer term, conditional dependencies should be indexed so only affected rows or sections need to be reevaluated.

Relevant implementation: [`UI.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.cs#L340-L416)

### 4. `ShowIf` evaluates only one representative material during multi-material editing

**Severity: High**

Enum selection explicitly scans all matching materials to detect a mixed value. In contrast, `ShowIf` resolves its source value from only `context.Material`.

If several selected materials have different values for the controlling property, conditional visibility is therefore decided by one representative material. A dependent row may be hidden even though some selected materials satisfy the condition, or shown even though some do not.

The expected mixed-value behavior should be defined explicitly. Possible policies include:

- Show when any selected material satisfies the condition.
- Show only when all selected materials satisfy it.
- Show the row with a mixed/partial-applicability indicator.

Relevant implementation: [`UI.MaterialSectionPresenter.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.MaterialSectionPresenter.cs#L224-L345)

### 5. Hidden Advanced properties can cause an empty 70 px mode button to appear

**Severity: High**

`HasAdvancedProperties()` checks the UI level, material property availability, and blacklist status, but it does not check `property.Hidden`. It also does not determine whether any Advanced property can ultimately be presented.

As a result, a shader containing only hidden Advanced properties can still display the Basic/Advanced button. The button then consumes horizontal space even though switching modes may reveal nothing.

The button-visibility policy should use the same effective-property filtering rules as the property presentation pipeline, or at least exclude hidden definitions.

Relevant implementation: [`UI.MaterialSectionPresenter.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.MaterialSectionPresenter.cs#L210-L222)

### 6. New row types bypass the existing responsive row-layout system

**Severity: High**

`RowLayoutCatalog` currently controls only a subset of rows, including Render Queue, Offset/Scale, Color, and Float. The new Shader mode, Enum, and float-backed Boolean controls still rely directly on fixed `SetWidth()` values.

The Shader row now reserves approximately:

```text
Categories button      20 px
Timeline button        20 px
Basic/Advanced button  70 px
Shader dropdown       316 px
Reset button           40 px
Padding                 2 px
--------------------------------
Fixed minimum         ~468 px
```

This leaves no width for the shader label until the available row width exceeds roughly 468 px. Below that point, the flexible label collapses first and the fixed controls can overflow or overlap.

Enum and Boolean rows similarly retain a fixed 316 px editor area. All row types should participate in one layout contract that defines:

- Minimum label width.
- Shrinkable editor width.
- Control hiding or compaction priority.
- Narrow-window breakpoints.
- A supported minimum main-panel width.

Relevant implementations: [`UI.RowViewFactory.MaterialShader.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.RowViewFactory.MaterialShader.cs) and [`UI.RowViewFactory.FloatKeyword.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.RowViewFactory.FloatKeyword.cs)

### 7. Dropdown typography changes are global rather than specific to the new Enum editor

**Severity: High**

PR #401 changes the shared dropdown styling used by Shader, Renderer, Enum, and other existing dropdowns. It applies:

- A maximum font size of 16.
- A minimum font size of 12.
- Best-fit resizing.
- Horizontal wrapping.
- Vertical truncation.
- A vertical text inset.

This changes existing UI globally even though the feature primarily needs an Enum editor. The combination of wrapping, a fixed single-row height, and vertical truncation is particularly fragile: text can attempt to wrap onto a second line that is immediately clipped, and adjacent dropdowns can display inconsistent font sizes because each one best-fits independently.

Dropdown styling should be split into explicit roles, for example:

```text
ShaderDropdown
CompactDropdown
EnumDropdown
DropdownPopupItem
```

Relevant implementation: [`UI.StyleSystem.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.StyleSystem.cs#L207-L241)

### 8. The Advanced accent geometry can overlap the existing property-label inset

**Severity: Medium-High**

Property rows currently use approximately:

- 1 px normal row padding.
- 3 px property-label inset.
- A resulting label start near 4 px.

The new Advanced accent is 5 px wide and is moved to the last sibling position, causing it to render over the row contents. Its geometry can therefore cover the first logical pixel of the label area. Whether a glyph is visibly clipped depends on font geometry, but the layout does not provide a dedicated accent gutter.

The property content inset should explicitly include the accent width and a gap, for example:

```text
Accent width + accent gap + normal padding
5 px         + 2 px       + 1 px = 8 px
```

Relevant implementation: [`UI.RowViewFactory.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.RowViewFactory.cs#L29-L52)

### 9. Every pooled virtual row contains every supported control type

**Severity: Medium-High**

`RowViewFactory.CreateTemplate()` creates all Renderer, Material, Shader, Texture, Color, Float, Keyword, Enum, and Boolean panels in one universal row template. `VirtualList` then clones that complete template for every cached visible row.

Consequently, even a simple Float row carries inactive Enum dropdown objects, Boolean controls, Shader controls, and every other editor panel. Adding each new editor type increases the GameObject and component count of every pooled row.

This may be acceptable at the current scale, but it creates a clear scalability limit. Separate pools or templates per row family would allow new editors to be added without increasing the cost of every unrelated row.

Relevant implementation: [`UI.RowViewFactory.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.RowViewFactory.cs)

### 10. Enum row construction can repeatedly scan all renderers and materials

**Severity: Medium**

Each Enum property calls `GetFloatValues()`, which traverses renderers and materials to collect same-named material values. With multiple Enum properties, the approximate cost becomes:

```text
Enum properties x renderers x materials
```

This work is repeated during full presentation refreshes and conditional refreshes. A presentation build should create one material-group snapshot or lookup cache and share it across all property descriptors.

Relevant implementation: [`UI.PropertyDescriptor.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.PropertyDescriptor.cs#L402)

### 11. Basic/Advanced state is scoped only by shader name

**Severity: Medium**

The Basic/Advanced state is stored as a `HashSet<string>` keyed only by `shaderName`. Different items, manifests, or zipmods that use the same shader name therefore share one session mode.

This may be intentional, but it should be an explicit product decision rather than an accidental consequence of the dictionary key. The intended scope should be clarified as one of:

- Global per shader name.
- Per manifest and shader.
- Per material.
- Per Material Editor window or selection.

Relevant implementation: [`UI.ShaderUiMode.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/UI/UI.ShaderUiMode.cs#L18-L38)

### 12. Global fallback definitions remain load-order dependent

**Severity: Medium**

`ShaderPropertyFallbackPolicy.MergeInto()` replaces a fallback property by name using the last loaded definition. If several manifests declare the same property name with different types, ranges, or visibility settings, the resulting legacy fallback depends on zipmod/manifest load order.

Fallback copies also reset `DeclarationOrder` to zero. Their final ordering then relies on dictionary enumeration and insertion history rather than a meaningful declaration order.

The fallback policy should define deterministic precedence, provenance, and conflict handling. At minimum, conflicts should be logged with both source identifiers rather than silently replacing the previous definition.

Relevant implementation: [`ShaderPropertyFallbackPolicy.cs`](https://github.com/IllusionMods/KK_Plugins/blob/41a661923b5e3912e5bdf7bb619972ca1f124979/src/MaterialEditor.Base/ShaderPropertyFallbackPolicy.cs#L47-L58)

## Suggested Remediation Order

1. Fix the fixed-width layout, Hidden Advanced button condition, and global dropdown styling.
2. Define multi-material `ShowIf` semantics and avoid disruptive full-list refreshes.
3. Move manifest models, parsing, conditions, and validation out of the UI layer.
4. Unify built-in and extension editors behind one editor strategy registry.
5. Cache material-group values during presentation construction.
6. Split the universal row template into row-family-specific pools if profiling confirms meaningful overhead.
7. Define explicit Basic/Advanced state scope and deterministic fallback precedence.

tomTom1010EEE:
# For PR #402

I reviewed the current PR head. The separate `Cubemap` property type and its dedicated persistence/reset paths are good design choices, and I did not find the broad layout regressions seen in the previous UI changes. However, I believe the following issues should be addressed or explicitly validated before merging.

## 1. Cubemap import performs expensive work on the Unity main thread

The file-dialog callback is correctly marshalled back to the main thread, but `yield return null` only postpones the operation by one frame. File reading, PNG decoding, projection, sampling, and Cubemap creation still execute synchronously on the main thread.

At the maximum supported face size, the conversion processes more than six million output pixels using projection math and bilinear sampling. This can visibly freeze Maker or Studio.

Please consider:

- Performing file I/O, hashing, and pure projection work on a worker thread.
- Restricting Unity object creation and pixel upload to the main thread.
- Alternatively, splitting conversion across multiple coroutine frames and displaying a processing state.

## 2. Peak memory usage needs profiling and tighter control

The source-file limits do not directly limit peak conversion memory. Import can simultaneously retain:

- Encoded PNG data.
- The decoded `Texture2D`.
- Source pixel arrays.
- A face conversion buffer.
- The resulting Cubemap and mipmaps.

Export additionally allocates six `Color[]` face buffers, a 2:1 panorama buffer, an output `Texture2D`, and PNG encoder data. At the maximum supported dimensions, peak usage may reach hundreds of MiB.

Please profile this on the actual Unity 5.6 game targets and consider:

- Reusing a single face buffer.
- Using `Color32` where possible.
- Processing faces incrementally.
- Reducing or making the maximum face size configurable.
- Adding an explicit conversion-memory budget.

## 3. The non-readable Cubemap GPU export path needs runtime validation

For non-readable Cubemaps, the implementation creates a temporary camera, Skybox material, RenderTexture, and readback operation for every face.

This introduces several risks:

- Six synchronous GPU readbacks may cause a noticeable stall.
- Face orientation may differ from the CPU-readable path.
- Gamma/linear color handling is not explicit.
- Results may vary according to the active project color-space configuration.

Please test this with a direction-labelled, color-coded Cubemap and compare:

- CPU-readable export.
- GPU-readback export.
- Export followed by re-import.
- Results across supported Maker and Studio targets.

It would also be preferable to reuse the temporary GPU resources across all six faces and explicitly select the intended read/write color space.

## 4. Original Cubemap remapping relies on material traversal order

`MaterialCubemapOriginalSnapshot.TryRemapToCurrentMaterials()` maps saved original values to current materials by list index whenever the material counts match.

This cannot detect a renderer/material ordering change. If traversal order changes after duplication or reload while the count remains the same, an original Cubemap may be restored to the wrong same-named material.

Please use a stable material identity where possible, such as a combination of:

- Renderer hierarchy path.
- Renderer/material slot index.
- Formatted material name.

A runtime test should also duplicate or reload an object after deliberately changing material traversal order.

## 5. Existing Texture2D import state is now updated optimistically

The revised file-selection flow immediately sets:

```csharp
textureItem.Changed = true;
textureItem.Exists = true;
```

However, character and Studio Texture2D imports are applied by their controllers during a later update. If reading or decoding the selected image fails, the UI can still temporarily report a successful import.

Please make the controller/repository expose the actual result, or refresh the row from the real material and edit state after the deferred import finishes. This is especially important because it changes an existing Texture2D path in a Cubemap-focused PR.

## 6. `CubemapConversion.cs` contains too many responsibilities

The new file is approximately 600 lines and currently owns:

- Content-addressed caching.
- Reference-counted lease lifetime.
- PNG validation and decoding.
- Panorama-to-Cubemap conversion.
- Cubemap-to-panorama export.
- CPU-readable extraction.
- GPU readback.
- Sampling, filtering, and face-orientation helpers.

Although it is not one giant method, it combines several independent failure and lifecycle domains. Please consider splitting it into focused components such as:

- `CubemapCache`
- `CubemapImportPipeline`
- `CubemapExportPipeline`
- `CubemapGpuReadback`
- `CubemapSampler`

`CubemapProjection.cs` is comparatively cohesive as a pure mathematical component and can remain separate.

## 7. Character and Studio implementations duplicate substantial Cubemap logic

The PR introduces parallel Character and Studio implementations for:

- Cubemap edit operations.
- Persistence models.
- Lease acquisition and release.
- Cache cleanup.
- Copy/import transformations.
- Original-value snapshot handling.

The implementations have already started to drift. For example, the Character `NullCheck()` validates `TexID`, `Property`, and `MaterialName`, while the Studio version only checks `TexID`.

The public persistence models may remain environment-specific, but the common Cubemap storage, lease, snapshot, and lifecycle operations should be moved into a shared internal service to reduce future divergence.

## 8. `UI.cs` is regaining infrastructure responsibilities

This PR adds more file-dialog, thread-marshalling, coroutine, import/export, watcher, file-writing, and state-refresh logic directly to `UI.cs`.

That moves the class back toward being a central implementation point after the recent UI modularization work. Please consider extracting components such as:

- `FileSelectionCoordinator`
- `TextureImportCoordinator`
- `CubemapImportExportCoordinator`

The main UI class should ideally coordinate presentation rather than own file processing and threading details.

## 9. Full Cubemap conversion currently runs while holding the global cache lock

`MaterialEditorCubemapCache.TryAcquire()` holds `lock (Sync)` while performing PNG decoding and the complete Cubemap conversion.

For a large input, this may hold the global lock for several seconds. It also serializes all cache users around the most expensive part of the pipeline.

Please limit the lock to short cache operations:

1. Check for an existing entry.
2. Perform conversion outside the lock.
3. Re-enter the lock to publish or deduplicate the result.
4. Update reference counts under the lock.

Unity API calls still need appropriate main-thread handling, but the cache lock should not remain held throughout the conversion.

---

The current CI build passing is a good baseline, but this feature also needs runtime coverage for performance, memory usage, face orientation, readable versus non-readable export, duplication/reset behavior, and card/coordinate/Studio persistence before merge.

PR #410, opened 2026-09-13
[ME] Fix character edit saving and local textures (2.6/7)

tomTom1010EEE:
## Description
Fix character cards losing edits when they contain only material renames or projector changes.
Fix local texture saving ignoring the selected mode and embedding texture bytes. Local mode now writes external files and stores only hash references; write failures do not silently fall back to bundled saving.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)


src/MaterialEditor.Core/Core.MaterialEditor.TextureSaveHandler.cs
```diff
@@ -68,11 +79,64 @@ internal static TextureContainer CreateTextureContainer(byte[] data)
...
+                    try
+                    {
+                        File.WriteAllBytes(temporaryPath, bytes);
+                        File.Move(temporaryPath, path);
+                    }
+                    finally
+                    {
+                        if (File.Exists(temporaryPath))
+                            File.Delete(temporaryPath);
+                    }
```
starstormhun, line 164: This shouldn't fail silently when writing the file fails. This implementation adds the hash to the save data even when it fails to write the file and doesn't notify the user.

src/MaterialEditor.Core/Core.MaterialEditor.TextureSaveHandler.cs
```diff
@@ -68,11 +79,64 @@ internal static TextureContainer CreateTextureContainer(byte[] data)
...
+                saveLocal = SceneSaveType?.GetValue(null, null)?.ToString() == "Local";
+            else
+#endif
+                saveLocal = KKAPI.Maker.MakerAPI.InsideMaker
+                    && CardSaveType?.GetValue(null, null)?.ToString() == "Local";
+
+            if (saveLocal)
+                SaveLocal(pluginData, key, data);
+            else
+                SaveBundled(pluginData, key, data, isCharaController);
```
starstormhun: Still seems to be missing SaveDeduped from the original implementation. (Studio-only)

src/MaterialEditor.Core/Core.MaterialEditor.TextureSaveHandler.cs
```diff
@@ -9,7 +9,7 @@ namespace KK_Plugins.MaterialEditor
 {
     /// <summary>
     /// Reads local and deduplicated version-2 texture data.
-    /// Writes the bundled version-1 format supported by the installed runtime.
+    /// Writes bundled version-1 or local version-2 data according to the save mode.
     /// </summary>
     internal sealed class TextureSaveHandler
```
starstormhun: Why isn't this extending KKAPI's TextureSaveHandlerBase anymore???
Don't unnecessarily limit its functionality for the sake of keeping the KKAPI requirement low, it makes no sense. Restore the original implementation please.

starstormhun: It should extend the original base class and register itself with KKAPI just like it used to, otherwise KKAPI won't know of ME and can't audit its files for example.

starstormhun: It also won't even display the save type selectors unless a plugin using its base class registers itself, meaning that if ME is the only installed plugin that wants to use the save type but isn't actually doing so properly, it will not work.

starstormhun (changes requested):
Needs fixes. See comments.
Also please make sure that the hash generating function for the local/deduped saving wasn't accidentally refactored, because if that changed in any way it should absolutely be restored, otherwise it can't be guaranteed to produce the same results.

src/MaterialEditor.Core/Core.MaterialEditor.TextureSaveHandler.cs
tomTom1010EEE: Agree, there's no need to keeping KKAPI requirement low

starstormhun (approved):

